PR #6, opened 2021-06-30
[Bugfix] new_filename injection

castingItAway:
When new_filename has an apostrophe, executing sqlite_query throws a syntax error or executes whatever is after it when possible. 
This fixes it by escaping apostrophes on the filename.

### DIFF
--- a/plugins/renamerOnUpdate/renamerOnUpdate.py
+++ b/plugins/renamerOnUpdate/renamerOnUpdate.py
@@ -461,6 +461,8 @@ except sqlite3.Error as error:
     sys.exit(1)
 
 # Looking for duplicate filename
+new_filename = new_filename.replace("'", "''")  # Fixing a bug whereby filenames with apostrophes would end the string prematurely, and would throw a syntax error
+                                                #when executing the query
 sqlite_query = "SELECT id FROM scenes WHERE path LIKE '%{}%';".format(new_filename)
 cursor.execute(sqlite_query)
 dupl_check = cursor.fetchall()


plugins/renamerOnUpdate/renamerOnUpdate.py
```diff
@@ -461,6 +461,8 @@ def makeFilename(scene_information, query):
     sys.exit(1)
 
 # Looking for duplicate filename
+new_filename = new_filename.replace("'", "''")  # Fixing a bug whereby filenames with apostrophes would end the string prematurely, and would throw a syntax error
+                                                #when executing the query
 sqlite_query = "SELECT id FROM scenes WHERE path LIKE '%{}%';".format(new_filename)
 cursor.execute(sqlite_query)
```
peolic, line 467: probably the [safer method](https://docs.python.org/3/library/sqlite3.html#sqlite3-placeholders)
```suggestion
sqlite_query = "SELECT id FROM scenes WHERE path LIKE ?;"
sqlite_args = ("%{}%".format(new_filename), )
cursor.execute(sqlite_query, sqlite_args)
```

Belleyy:
I spotted this error too, I think i fixed it with:
```py
cursor.execute("SELECT id FROM scenes WHERE path LIKE ?;", ["%" + new_filename])
```

peolic:
@Belleyy either one should work, just FYI that you changed the `LIKE` argument a little bit in your comment.

Belleyy:
> 
> 
> @Belleyy either one should work, just FYI - you changed the `LIKE` argument a little bit in your comment.

Yea, the last % is not really needed since we check if a filename with same name exist, we only need to check if a path end with the `new_filename`

PR #10, opened 2021-07-15
[Plugin] Github scraper checker

Belleyy:
I am pretty sure that this plugins could work with a stash version < v0.8.
Require **Request** module 

- Status Check

Show in log if you miss a scraper file, or a new update of this scraper exist.

- Getting new files

Only get file that don't exist in your scraper folder.

- Overwrite everything

Commented in .yml because i don't want to miss-click on this button.
It will overwrite all your scrapers files, will remove personal things like API Key.


jeremymeyers:
Perhaps something in the YML for filenames to exclude from updates (i.e. those with API info etc)

PR #16, opened 2021-09-19
Add plugin which parses scene and gallery filenames on add

gitgiggety:
Add a plugin which tries to parse the names of scenes and galleries when they're added / scanned. The primarily supported format is `{studio}.{year}.{month}.{day}.{performer}.{performer}.{title}`. Studio can optionally include ".com". Some "parts" are ignored like resolution, some release group names etc. Performers can be separated by an optional "and" which will be ignored as well. Performers are both searched based on name and aliases, for which aliases are split based on `,` and `;` and needs to result in an "exact" match meaning "First Last" won't match a performer which name or alias is just "First". When there are multiple
matches (for example when filename is "first1.first2.last2" and multiple performers having the "first1" alias) all matched performers will be added.


gitgiggety:
I've tested this by removing all scenes and galleries from the database and rescanning the files. Of the 460 and a bit scenes I have there are only about 20 to 30 which "fail" (i.e.: can't be parsed or are parsed incorrectly). But most of them are because they have a very different filename format (and in my case: are really old downloads, all "new" downloads follow said format). And some fail because the filename only contains a firstname and multiple performers having an exact match (mostly on alias) for that firstname which means I just add all matched performers.

PR #30, opened 2021-12-31
[renamerOnUpdate] config.py corrections

PhuriousGeorge:
minor corrections

### DIFF
--- a/plugins/renamerOnUpdate/config.py
+++ b/plugins/renamerOnUpdate/config.py
@@ -43,7 +43,7 @@ default_template = "$date $title"
 # File to save what is renamed, can be useful if you need to revert changes.
 # Will look like: IDSCENE|OLD_PATH|NEW_PATH
 # Leave Blank ("") or use None if you don't want to use a log file, or a working path like: C:\Users\USERNAME\.stash\plugins\Hooks\rename_log.txt
-log_file = r""
+log_file = ""
 
 ######################################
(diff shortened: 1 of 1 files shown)


plugins/renamerOnUpdate/config.py
```diff
@@ -43,7 +43,7 @@
 # File to save what is renamed, can be useful if you need to revert changes.
 # Will look like: IDSCENE|OLD_PATH|NEW_PATH
 # Leave Blank ("") or use None if you don't want to use a log file, or a working path like: C:\Users\USERNAME\.stash\plugins\Hooks\rename_log.txt
-log_file = r""
+log_file = ""
```
Belleyy, line 46: I don't agreed with this change. If you copy directly the path and not having the **r**aw before the quote would give a error.
```py
log_file = r"C:\Users\Winter\Downloads\Video\vid.mp4"
log_file2 = "C:\Users\Winter\Downloads\Video\vid.mp4"
print(log_file)
print(log_file2)
#SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
```
The user would need to escape every `\`. Maybe if you don't use Windows you don't really care but for user like me it's easier. 
So unless you have a good reason to remove it, i would keep it.

PhuriousGeorge:
@Belleyy Good call my bad.

PR #33, opened 2022-01-10
As per discord, merging into one file until the length requires addit…

jeremymeyers:
…ional

https://discordapp.com/channels/559159668438728723/645283324071247911/930168290670706749


bnkai:
As mentioned in discord merging this for now ? When the list or readme gets longer we can re split.

Issue #28, opened 2021-12-05
[renamerOnUpdate] Deletes characters that are in the template


coscosname:
Deletes characters that are in the template if one of the values is empty.
e.g.
```python
filename_template = "[$studio] $date _ $title"
scene_information["studio"] = "studio1"
scene_information["date"] = ""
scene_information["title"] = "title1"
```

In this case we get
```
[studio1] title1
```
instead of
```
[studio1] - title1
```
This applies to characters
```
- and _
```

https://github.com/stashapp/CommunityScripts/blob/975d179d8b02feeb351edc5658abd42e5b7a0bd7/plugins/renamerOnUpdate/renamerOnUpdate.py#L182

bnkai:
Relevant PR merged. Closing.

PR #45, opened 2022-05-21
[renamerOnUpdate] Associate file, On/Off Hook, more option

Belleyy:
## Feature
- Adding Plugin Task to disable/enable the hook
image
- Rename file associated to the video file (Subtitle, funscript) #46

## Option
- ⚠️ Changed `performer_ignore_male` to `performer_ignoreGender` so you can choose other gender as well.
- Add option to put your filename in lowercase
- Add option to replace space from Stash field (title, performer, studio, ...)
- Add option to remove specific character from filename, By default `,#` ,these character was before in the regex to remove illegal character in windows. User didn't know why these character wasn't on the filename so it should be more clear now.
- Add option to remove prepositions from the beginning of titles.

## Misc
- ❗ ⚠️ ❗ Changed the regex to create the filename, it should be better but report if it's not correct. #28 
- ⚠️ Audio & Video codec always in uppercase
- Changed the logic of performer count, now filter then count.
- Adding a variant of Apostrophe (`”“`) to be replaced
- Put the url of your stash in the config
- Adding a DRY-RUN mode for testing purpose

### DIFF
--- a/plugins/renamerOnUpdate/renamerOnUpdate.py
+++ b/plugins/renamerOnUpdate/renamerOnUpdate.py
@@ -304,7 +304,7 @@ FILENAME_REMOVECHARACTER = config.removecharac_Filename
 
 PERFORMER_SPLITCHAR = config.performer_splitchar
 PERFORMER_LIMIT = config.performer_limit
-PERFORMER_IGNORE_MALE = config.performer_ignore_male
+PERFORMER_IGNOREGENDER = config.performer_ignoreGender
 PREVENT_TITLE_PERF = config.prevent_title_performer
 
 SQUEEZE_STUDIO_NAMES = config.squeeze_studio_names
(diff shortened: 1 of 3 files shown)


plugins/renamerOnUpdate/renamerOnUpdate.py
```diff
@@ -275,15 +291,22 @@ def exit_plugin(msg=None, err=None):
...
+
+FIELD_WHITESPACE_SEP = config.field_whitespaceSeperator
 
+FILENAME_LOWER = config.lowercase_Filename
 FILENAME_SPLITCHAR = config.filename_splitchar
+FILENAME_REMOVECHARACTER = config.removecharac_Filename
 
 PERFORMER_SPLITCHAR = config.performer_splitchar
 PERFORMER_LIMIT = config.performer_limit
 PERFORMER_IGNORE_MALE = config.performer_ignore_male
```
DogmaDragon, line 307: Should be 
`PERFORMER_IGNOREGENDER = config.performer_ignoreGender`

Belleyy:
I changed the way to create the filename, it should be better but i could be wrong too so report if it's wrong.

PR #47, opened 2022-05-24
Adding requested option to create collections from ALL scene tags

Darklyter:


Darklyter:
Honestly I'm not sure why all of those old commits are showing since they're already PRed and merged, but I'm not sure how to clear them out so figured I'd go ahead and toss it up for the 7 or so lines of difference.

PR #62, opened 2022-07-14
add and update plex theme

halorrr:
Will be moving themes here starting with the Plex one, as per https://github.com/stashapp/stash/issues/2741#issuecomment-1183786833

I've applied the updates that came up in that issue to the theme as well.

Went with the approach of each theme having its own folder so that a `README` file can be present and automatically visible when in the folder for the theme.


WithoutPants:
Please also add a theme section to the root readme file.

halorrr:
Added the theme section to the main readme

PR #67, opened 2022-07-26
[renamerOnUpdate] Add option to replace thing

Belleyy:
Add a option to replace thing in your filename.

The second element of the list determine the system used. If you don't put this element, the default is word
`regex`: match a regex, `word`: match a word, `any`: match anything
difference between `word` & `any`: word is between seperator (`space  _  -`), any is anything ('ring' would replace 'during')

## Exemple:   
`"Scene": ["Sc.", "word"]`    - Replace Scene by Sc.
`r"S\d+:E\d+": ["", "regex"]` - Remove Sxx:Ex (x is a digit)
```py
replace_words = {
    r"S\d+:E\d+": ["", "regex"],
    "Be": ["lol", "word"],
    "Do": ["Is", "any"]
}
#Darcy Does It Best - S22:E6 -> Darcy Ises It Best
```

Also fix a bug when trying to disable/enable hook

### DIFF
--- a/plugins/renamerOnUpdate/config.py
+++ b/plugins/renamerOnUpdate/config.py
@@ -82,6 +82,15 @@ filename_splitchar = " "
 # replace space for stash field (title, performer...), if you have a title 'I love Stash' it can become 'I_love_Stash'
 field_whitespaceSeperator = ""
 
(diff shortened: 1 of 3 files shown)


Belleyy:
included in #70

Issue #68, opened 2022-07-26
[renamerOnUpdate] Advanced filename formatting


coscosname:
Please add more formatting options like [Advanced formatting in Calibre - Conditionally including text](https://manual.calibre-ebook.com/template_lang.html#advanced-formatting)

Example:
I want to use the `"[$studio] $date - $title"`  template but the scene doesn't have a date specified so I get `[studio] - title.mp4` file name

I would rather get the name `[studio] title.mp4`

bnkai:
Closing as completed

Issue #66, opened 2022-07-23
[renamerOnUpdate] Add plugin button to rename all files


JediMaster25:
Add a plugin button to execute the plugin for all scenes, and change the filenames same as the Sqlite_Renamer script. This plugin offers more options for renaming the title than the script.

bnkai:
Closing as completed

PR #70, opened 2022-07-28
[renamerOnUpdate] Major Update (Bulk, Mover...)

Belleyy:
## Features

- Template to move your files 
- Bulk task to rename all your scenes with a button #66 
- Dry-run (Do a trial run with no permanent changes)

## New option
- More field (mostly movie related)
- option to sort of the order of performer (Name, ID, Favorite, Rating)
- option to keep performer if it reach the performer limit
- group in template, you can group character with `{}`. kinda #68
  Exemple: `[$studio] {$date -} $title` The `-` is with the date, so if the date is empty, `-` will be remove as well
- Add a option to replace thing in your filename. (#67)

  The second element of the list determine the system used. If you don't put this element, the default is word
`regex`: match a regex, `word`: match a word, `any`: match anything
difference between `word` & `any`: word is between seperator (`space  _  -`), any is anything ('ring' would replace 'during')

  Exemple:   
`"Scene": ["Sc.", "word"]`    - Replace Scene by Sc.
`r"S\d+:E\d+": ["", "regex"]` - Remove Sxx:Ex (x is a digit)
```py
replace_words = {
    r"S\d+:E\d+": ["", "regex"],
    "Be": ["lol", "word"],
    "Do": ["Is", "any"]
}
#Darcy Does It Best - S22:E6 -> Darcy Ises It Best
```

## Deprecation
- Remove field to have the correct length if reached the max .

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This list keeps track of scripts and plugins in this repository. Please ensure t
 Category|Triggers|Plugin Name|Description|Minimum Stash version
 --------|-----------|-----------|-----------|---------------------
 Scraper|Task|[GHScraper_Checker](plugins/GHScraper_Checker)|Compare local file against github file from the community scraper repo.|v0.8
-Maintenance|Scene.Update|[renamerOnUpdate](plugins/renamerOnUpdate)|Rename your file based on Stash metadata.|v0.7
+Maintenance|Task<br />Scene.Update|[renamerOnUpdate](plugins/renamerOnUpdate)|Rename/Move your file based on Stash metadata.|v0.7
 Maintenance|Set Scene Cover|[setSceneCoverFromFile](plugins/setSceneCoverFromFile)|Searchs Stash for Scenes with a cover image in the same folder and sets the cover image in stash to that image|v0.7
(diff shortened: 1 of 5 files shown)


scruffynerf:
Found a bug, replicated repeatedly:

Move a file so it's not where Stash expects, but it's still within the scan.
Have another file already in place, which this plugin would want to use as the name/location for the first file when it renames it.  (so the renamer will fail to change the name/path, and leave the file alone)

When Renamer attempts to move/rename the file, it will fail with 'database is locked' error.
The file path isn't updated, and so the file is considered lost (it didn't save at the rename location NOR did it record the new location), so a clean will now show the file entry in the DB as 'missing and removable' and no matter if you scan again, it won't add, because Renamer fails it again.  Without Renamer, the scan would work, it's Renamer causing the bug.

bnkai:
@Belleyy merging this as it looks ok. Can you update when possible to make sure it doesnt run with the filles-refactor version (at least till you add support for it) ? You can use the `systemStatus` query to get the database schema version as an int <https://github.com/stashapp/stash/wiki/API#get-system-status>

PR #77, opened 2022-09-25
titleDiff plugin

o-fl0w:
Plugin with a task. When ran tags all scenes where title differs from filename (with or without extension).

### DIFF
--- a/plugins/titleDiff/titleDiff.js
+++ b/plugins/titleDiff/titleDiff.js
@@ -1,0 +1,73 @@
+var TAG_NAME = "_titleDiff" // Name of tag to set when scene title differs from filename, will create if not found
+
+
+function main() {
+    switch (input.Args.mode) {
+        case "diff":
+            diff()
+    }
+}
+
+function diff() {
+    log.Info("diff: Begin diff title/filename")
+    var sceneIds = findDiffingScenes()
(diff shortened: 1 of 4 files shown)


plugins/titleDiff/titleDiff.js
```diff
@@ -0,0 +1,73 @@
...
+    var result = gql.Do(query)
+    if (result.findScenes == null) {
+        log.Info("diff: No scenes found")
+        return
+    }
+    var scenes = result.findScenes.scenes
+    var diffs = []
+    for (var i = 0; i < scenes.length; i++) {
+        var filename = scenes[i].path.split(/.*[\/|\\]/)[1];
+        var filenameNoExt = filename.replace(/\.[^/.]+$/, "")
```
bnkai, line 31: Can you add an extra check for future proofing here?
With the <https://github.com/stashapp/stash/pull/3006> PR there can be scenes with no files. For those the path is empty and thus you get `Plugin returned error: TypeError: Cannot access member 'replace' of undefined ` and the plugin exits. For those cases i think just skipping the scene is ok

scruffynerf:
@o-fl0w , you still around?

DogmaDragon:
Abandoned.

PR #83, opened 2022-10-12
Timestamp.trade

Tweeticoats:
I've created the api https://timestamp.trade to sync markers between stash instances and xbvr.
This sits along side other metadata databases like stashdb while we wait for the feature to be added there.

### DIFF
--- a/plugins/timestampTrade/timestampTrade.py
+++ b/plugins/timestampTrade/timestampTrade.py
@@ -32,8 +32,9 @@ def processScene(s):
 
 
 def processAll():
-
+    log.info('Getting scene count')
     count=stash.find_scenes(f={"stash_id":{"value":"","modifier":"NOT_NULL"},"has_markers":"false"},filter={"per_page": 1},get_count=True)[0]
+    log.info(str(count)+' scenes to submit.')
     for r in range(1,int(count/per_page)+1):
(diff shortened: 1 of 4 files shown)


plugins/timestampTrade/timestampTrade.py
```diff
@@ -0,0 +1,85 @@
...
+import sys
+import requests
+import json
+
+per_page = 100
+
+def processScene(s):
+    if len(s['stash_ids']) > 0:
+        for sid in s['stash_ids']:
+            #                print('looking up markers for stash id: '+sid['stash_id'])
```
bnkai, line 14: From your plugin description it seems that you want the stash_ids from `stashdb`? Don't you need to check the endpoint to make sure you dont get stash_ids from another stash-box instance? Not sure if there is another public/private stash-box instance except stashdb (maybe tpdb) but it doesnt hurt making sure

Tweeticoats: The database does keep track of stashid and endpoints on the api side but for simplification all endpoints will hit the same api.
I'm assuming as uuid are configured to  generate mostly randomly there will not be any collisions between two instances, while technically possible the amount of possible numbers it's like winning the lotto jackpot several times in a row.
The api does return the endpoint and stashid so the plugin can be updated to verify that the stashid does point to the right endpoint before accepting the results.

PR #89, opened 2022-10-28
[PathParser] Plugin

FauxMatt:
Path Parser shares some similarities with filenameParser and defaultDataForPath.

Unlike those examples, it allows more advanced matching capabilities against the file path for scenes and can update scene attributes based on sub-expression in the path.

For example, if there is a folder for a movie named "Movie Name (2010)" and the scene is split into multiple files "Scene Title - Scene [N].mp4" it can extract/update Movie Name and Scene Title.

It supports updating scenes tagged with [Run], testing rules against scenes tagged with [Test], and automatically updating scenes as they are scanned.


bnkai:
@FauxMatt can you update the main README and add to the existing list? <https://github.com/stashapp/CommunityScripts#plugins>
From a quick look it seems that it needs the latest version of stash  >= 0.17 (after the files refactor) ?

FauxMatt:
@bnkai updated the main README.

PR #92, opened 2022-11-02
[renamerOnUpdate] Perform a stricter check in config_edit to avoid breaking the config

hphpanon:
Because this check was using `in`, using the Dry Run setting from inside the Stash UI was destroying the `dry_run_append` config value (by making it another `dry_run` value), thereby breaking the config. As it turns out `dry_run` is both in `dry_run` **and** `dry_run_append`. `config_edit` appears to only be used in cases with strict config value matching so I don't expect this loose comparison is an expected behavior.

### DIFF
--- a/plugins/renamerOnUpdate/renamerOnUpdate.py
+++ b/plugins/renamerOnUpdate/renamerOnUpdate.py
@@ -342,7 +342,7 @@ def config_edit(name: str, state: bool):
         with open(config.__file__, 'w', encoding='utf8') as file_w:
             for line in config_lines:
                 if len(line.split("=")) > 1:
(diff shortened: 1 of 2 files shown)


hphpanon:
I believe a similar loose check is leading to double-logging the enable/disable flag, but I'm not familiar enough with the structure of `PLUGIN_ARGS` to immediately know how to address that and I'm more OK with double-logging than breaking the config (which breaks the plugin).
image

GloveNinja217:
I'm experiencing the same issue where after performing a dry run, the `dry_run_append` config value is being overwritten as `dry_run`

Issue #97, opened 2022-11-10
[renamerOnUpdate] Element feature requests


echo6ix:
Four element field feature requests for renamerOnUpdate:

1. _$filename_ - The incumbent filename of the file.
Example use-case: To keep the existing file name as-is, but change file path location only.
2. _$stashid_scene_ - StashID of the scene
3. _$stashid_performer_ - StashID of the performer(s)
4. _$studio_code_ - The new scene studio code added in recent develop versions

PR #103, opened 2022-11-18
[renamerOnUpdate] more field

Belleyy:
some from #97

### DIFF
--- a/plugins/renamerOnUpdate/renamerOnUpdate.py
+++ b/plugins/renamerOnUpdate/renamerOnUpdate.py
@@ -107,25 +107,33 @@ def graphql_getScene(scene_id):
         title
         date
         rating
+        stash_ids {
+            endpoint
+            stash_id
+        }
         organized""" + FILE_QUERY + """
         studio {
-          id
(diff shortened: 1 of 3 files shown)


plugins/renamerOnUpdate/renamerOnUpdate.py
```diff
@@ -450,7 +466,12 @@ def extract_info(scene: dict, template: None):
     scene_information['current_directory'] = os.path.dirname(scene_information['current_path'])
     scene_information['oshash'] = scene['oshash']
     scene_information['checksum'] = scene.get("checksum")
+    scene_information['studio_code'] = scene.get("code")
 
+    if scene.get("stash_ids"):
+        #todo support other db that stashdb ?
+        scene_information['stashid_scene'] = scene['stash_ids'][0]["stash_id"]
```
bnkai, line 473: Haven't actually tested yet
If the scene doesnt have any stash_ids will the above work ok?
Won't an empty array pass the `if` and then fail due to `scene['stash_ids'][0]` ?

PR #112, opened 2022-11-30
[phashDuplicateTagger] Normalize tags on duplicate scene processing

SmutsigGulag:
Often there's more tags/info on a lower quality scene. This aims to normalize the tags between duplicate scenes, avoiding data loss.


DogmaDragon:
Plugin was moved to the diferent repository. See. https://github.com/stg-annon/StashScripts/tree/main/plugins/phashDuplicateTagger

Issue #111, opened 2022-11-26
[renamerOnUpdate] Error -> Cannot query field "code " on type "Scene"


BrekiTomasson:
I updated the plugin with the changes from 2 days ago and am now getting this error:

```
2022-11-26 11:43:30
Error   
Plugin returned error: exit status 1
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate] ConnectionError: GraphQL query failed: 422 - b'{"errors":[{"message":"Cannot query field \\"code\\" on type \\"Scene\\".","locations":[{"line":23,"column":13}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}},{"message":"Cannot query field \\"code\\" on type \\"Scene\\".","locations":[{"line":23,"column":13}],"extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}}],"data":null}'
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate]     raise ConnectionError(f"GraphQL query failed: {response.status_code} - {response.content}")
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate]   File "/root/.stash/plugins/renamerOnUpdate.py", line 93, in callGraphQL
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate]     result = callGraphQL(query, variables)
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate]   File "/root/.stash/plugins/renamerOnUpdate.py", line 212, in graphql_findScene
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate]     scenes = graphql_findScene(config.batch_number_scene, "ASC")
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate]   File "/root/.stash/plugins/renamerOnUpdate.py", line 1318, in <module>
2022-11-26 11:43:30
Error   
[Plugin / renamerOnUpdate] Traceback (most recent call last):
```

bnkai:
Can you check and post which stash version you are using?
The plugin checks that the version is after the file refactor PR but doesnt check for the PR that added the `code` section.
If you are using a dev version update to the latest

BrekiTomasson:
v0.17.2, Build hash: 02c2ad3f

bnkai:
Issue is resolved with latest commit 
Closing this

PR #113, opened 2022-11-30
[renamerOnUpdate] fix bug

Belleyy:
- Fix studio_code order.
- Fix if the user don't have the `code` migration #111


DogmaDragon:
It applies the $studio_code now but reverts because of the undefined FILE_REFACTOR. 

`[Plugin / renamerOnUpdate] error when trying to update the database (name 'FILE_REFACTOR' is not defined), revert the move...`

Replacing line 1151
`                if FILE_REFACTOR:`
with
`                if db_version >= 32:`
fixed it for me.

PR #135, opened 2023-01-05
add custom_javascript and open_scene_in_iina

clangmoyai:
* add custom_javascript folder
* add open_scene_in_iina


philpw99:
This is actually a great script ! I intent to make a little change and let the android has the same kind of feature.
You already did 90% of the job. Thank you !

DogmaDragon (changes requested):
Move the screenshot to `/scripts/open_scene_in_iina` directory.

PR #137, opened 2023-01-08
files for movieFromScene plugin.

philpw99:
Sorry I still have no clue how github fork works....


scruffynerf:
I like the idea of this, but not the custom gui.  The core concept of creating movies that are missing is good.

DogmaDragon (changes requested):
Needs to be rebased against latest Groups schema.

PR #138, opened 2023-01-09
Scene card video player

philpw99:
Add a "Play" button to the scene cards and allow you to quickly play videos directly from the scene list.
This is a custom Javascript used in Settings->Interface->Custom Javascript.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,3 @@ Category|Plugin Name|Description|Minimum Stash version
 --------|-----------|-----------|---------------------
 Kodi|[Kodi Helper](scripts/kodi-helper)|Generates `nfo` and `strm` for use with Kodi.|v0.7
 Maintenance|[Stash Sqlite Renamer](scripts/Sqlite_Renamer)|Renames your files using stash's metadata.|v0.7
-
-## Custom Javascripts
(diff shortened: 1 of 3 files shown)


bnkai:
IMHO if the script is to be updated mostly in your repo it is better to either leave only the README with the  link or nothing at all here (to make maintenance easier).
The url mentioned here <https://docs.stashapp.cc/add-ons/scripts#pwplayerjs---scene-card-quick-player> is ok?

scruffynerf:
https://github.com/philpw99/StashCustomJavascripts

Quiet since January... Some nifty scripts, but I'd really like to see activity before I add essentially a pointer to there.  We should definitely list the scripts in docs.stashapp.cc

Maybe turn things into a js plugin?

DogmaDragon:
Both are already listed on [docs.stashapp.cc](https://docs.stashapp.cc/add-ons/scripts)
https://docs.stashapp.cc/add-ons/scripts/#pwplayerjs---scene-card-quick-player
https://docs.stashapp.cc/add-ons/scripts/#blurrycardbackground

DogmaDragon:
https://docs.stashapp.cc/scripts/list/#pwplayerjs-scene-card-quick-player

PR #140, opened 2023-01-22
[renamerOnUpdate] handling of duplicate filenames of different scenes

AnonTester:
Adds duplicate handling of same filename of different scenes testing all duplicate_suffix config variables before aborting.


MikeLund:
If you're not using any filename renaming, this change will make it so your existing files will be renamed to `None_X` (X being next duplicate_suffix). This is probably because `template["filename"]` is empty in these cases.

MikeLund:
Using that, nothing gets appended to the original filename -- it just tries using the same (already-existing) filename until `len(DUPLICATE_SUFFIX)` is reached. 

Using `create_new_filename(scene_information, scene_information['current_filename'])`, it becomes the existing filename (including file extension) + DUPLICATE_SUFFIX + file extension. Probably a much cleaner way to do it. Thanks for your original change, btw!

PR #145, opened 2023-02-13
[renamerOnUpdate] Refactoring + cleanup

bnkai:
Refactoring of the code to make it easier to maintain
- No functionality was changed/removed and no config options either so it `should` work withought any changes to existing config
- Only new addition atm is making the bulk processing work by batches 

Renamer plugin and sqlite renamer script were removed as they were obsolete and all their functionality was moved to renamerOnUpdate anyway


scruffynerf:
Hopeful to use this as the basis for the RoU cleanup, prep for v24, and overall revamp of settings needed in plugins.

Issue #48, opened 2022-05-24
[markerTagToScene] Add all tags of Scene Marker to the Scene


IAmKontrast:
markerTagToScene v1.0 adds the primary tag of Scene Marker to the Scene on marker create/update.
I would like to have the option to activate 'all tags'.

scruffynerf:
Looks like we have 3 different approaches to solve this and hopefully we'll get one in place soon.
I've asked @stg-annon to review his version of this as well.

feederbox826:
closed by #311

PR #152, opened 2023-03-22
[markerTagToScene] Add all tags of Scene Marker to the Scene

IAmKontrast:
Option to add all scene marker tags to the scene.
Also transformed code from js to python.

Solves #48


scruffynerf:
Let's get some version of this in, whether it's this one or the one in @opensolutionbroadcast 's PR.

DogmaDragon (changes requested):
Needs a rebase.

PR #154, opened 2023-03-23
[plugin][markerTagToScene] Adding task option, enabling copy of ALL marker tags instead of only primary

opensolutionbroadcast:
Adding task to markerTagToScene plugin that operates on all markers. Added config options to markerTagToScene plugin to enable copying ALL marker tags to scene instead of only the primary tag.


DogmaDragon (changes requested):
Need to a rebase.

PR #155, opened 2023-03-28
[markerTagToScene] add all marker tags to scene not just primary

stg-annon:
looks at all marker tags from `SceneMarkerUpdateInput` and compares them to existing scene `tag_ids` adds any tags that are missing from scene


stg-annon:
so js plugins seem to have an issue with arrow functions as well as some other issues I'm unfamiliar with, I could rewrite this in python but I'm not sure that is the ideal solution here so I will draft a PF for it if it is not wanted feel free to close it

xiosensei:
following

bnkai:
@stg-annon sorry for the late review
how does this compare to #152, #154 ?

stg-annon:
@bnkai #154 might be preferred as is remains as JS plugin

This PR introduces a library requirement which might not be preferred but its written in such a way that I believe it is more maintainable

scruffynerf:
@stg-annon still feel the same way? This has sat a while and between this and #154 , i think one of these go in, or some hybrid approach.  Preference?

stg-annon:
this would be easy for me to maintain but I don't love introducing requirements where I don't have to, if `stashapp-tools` becomes a standard package I'd probably go with this

Issue #156, opened 2023-03-30
[timestampTrade] Incorrect JSON sent to timestamp.trade API


pornman1500:
Stash Version: v0.20.0

The JSON object sent to the timestamp.trade API through the HTTP POST method is not being updated correctly and results in the scene not posting to the online database.  Here is an example JSON that is being sent with REDACTED values removing sensitive information.

```
{
  "title": "March 2023 Flavor Of The Month Dakota Tyler - S3:E8",
  "code": "165812",
  "details": "Ricky Spanish is really, really into his stepsister Dakota Tyler, but she likes to make stupid bets. For example, she bets Ricky she can blow the biggest bubbles and chug a drink the quickest. At every turn, Dakota uses her tight petite body to her advantage to make sure she wins. Today, Dakota and Ricky are going through clothes to give away when Dakota bets Ricky she can fit in the suitcase. If she wins, Dakota claims she can suck Ricky's dick.\n\nRicky tries to tell Dakota to cut it out, but he takes her up on the bet anyway because he genuinely believes there's no way his tiny stepsis can fit into the suitcase. Dakota has just proved him wrong with her adopted mom, Charli Phoenix, comes in to see what's taking the stepsiblings so long. Dakota is hidden by the suitcase, which means she can pop Ricky's stiffie out and begin sucking it as he does his best to hold the moan. When Charli leaves, Ricky announces he shouldn't fuck Dakota but he's going to.\n\nFlipping Dakota around so she's on her knees in the suitcase, Ricky shoves it in. The stepsibs quickly get their clothes off and take the suitcase off the bed, then they get busy. Dakota rides Ricky in cowgirl, then sucks her juices from his hardon before remounting him in reverse cowgirl. Rolling onto her back, she spreads her thighs and begs Ricky to bang her pint sized pussy. Ricky has just delivered a big fat creampie when Charli returns and catches them in the aftermath of a good fuck.",
  "director": "",
  "url": "https://princesscum.com/video/watch/165812/march-2023-flavor-of-the-month-dakota-tyler-s3e8",
  "date": "2023-03-01",
  "rating100": null,
  "path": "REDACTED ",
  "interactive": false,
  "interactive_speed": null,
  "captions": null,
  "created_at": "2023-03-07T07:23:09Z",
  "updated_at": "2023-03-09T03:51:21Z",
  "file_mod_time": "2023-03-07T07:21:11Z",
  "last_played_at": null,
  "resume_time": 2.829424,
[...]

PR #159, opened 2023-04-16
Fixing bugs with the plugin with submission.

Tweeticoats:
Calling the stash graphql api manually as I only care about a few fields when submitting data.
Improved logging message and report % of files being processed.
Use a persistent session with timestamp.trade instead of connecting for each request.


pornman1500:
@StashAppDev This PR resolves #156.  Is there anything else that needs to be done before merging?

PR #160, opened 2023-04-22
Add a ComicRack Metadata Extractor for cbz based Stash Galleries

yoshnopa:
As the name proposes, this Plugin adds support for importing metadata from the ComicRack Standard for cbz files.
This data is saved in a ComicInfo.xml file in the archive on the top level, containing certain Tags with metadata.
As it is not exactly possible to match everything with stash 1:1, I added a yml file with a dictionary that matches the XML Tags with the variables of a stash gallery. 

The current values are what I use, but I've already added stuff into the code to be able to work with every value a stash gallery can have.
The default dictionary can be changed, probably the writer:studio matching is the most controversial. 

For more info on the standard, you may look to into
https://github.com/comictagger/comictagger

Also, right now, the data is always applied when updating a gallery, that means that you are de facto unable to set certain values yourself, as the update overwrites it with the comicinfo.xml data. I personally prefer this behaviour, but it might be of interest to change the hook to something less aggressive.


scruffynerf (approved):
Looks good.

PR #165, opened 2023-06-17
Added folder based gallery support

HijackHornet:
Updating my plugin to also consider folder based galleries


scruffynerf:
Neeed to test this...

PR #169, opened 2023-07-20
[renamerOnUpdate] fix handling of duplicate filenames of different sc…

AnonTester:
…enes with no renaming template followup of https://github.com/stashapp/CommunityScripts/pull/140


scruffynerf:
Will review and see if we still need this.   If so. Will add it in as part of revamp of RoU

mr-szgz:
> Will review and see if we still need this.   If so. Will add it in as part of revamp of RoU

a revamp? tell me more :)

PR #171, opened 2023-08-21
Stash userscript library

Teda1:
I've refactored @7dJx1qP/redsheep1's  Stash Userscript Library such that it can be used for plugin development directly into Stash. That library has made it easy for many users to build on, to create various user scripts. My thinking is that if the library was made available for plugin development, user scripts that could have been provided as a plugin could easily be transitioned into a plugin.

You must ensure the library is loaded before other scripts, so I have numbered the directories to enforce a load order. From there, I basically copied the content of the Stash Stats user script and provided it as its own plugin.

### DIFF
--- a/plugins/1. stashUserscriptLibrary/StashUserscriptLibrary.yml
+++ b/plugins/1. stashUserscriptLibrary/StashUserscriptLibrary.yml
@@ -1,0 +1,6 @@
+name: Stash Userscript Library
+description: Exports utility functions and a Stash class that emits events whenever a GQL response is received and whenenever a page navigation change is detected
+version: 1.0
+ui:
+  javascript:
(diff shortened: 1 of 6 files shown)


DogmaDragon:
This is cool. My only concern would be that making them part of plugins removes the flexibility that userscript managers offer.

Teda1:
There will eventually be an option to toggle plugins on and off the same you can with user scripts, which would improve plugin flexibility. That aside, this work doesn’t negate the ability to use user scripts. It simply provides a way to keep the extra added function in-house rather than managing them across browsers. With this implementation, porting the functionality back and forth from a plugin to user scripts or vice versa is as simple as copy and pasting.

PR #178, opened 2023-09-24
Add dupeMarker

feederbox826:
This plugin marks exact duplicate markers for manual removal by the user


scruffynerf (approved):
Looks good.

Issue #110, opened 2022-11-20
[Fix] setSceneCoverFromFile fails to run with SSL and self-signed certificate


Ksrx01:
I'm running Stash over SSL with a self-signed certificate, tried to use this plugin, which failed, as it obviously was unable to verify the certificate. So I change **Line:64** in `stash_interface.py` from:

> `response = requests.post(self.url, json=json, headers=self.headers, cookies=self.cookies)`

to

> `response = requests.post(self.url, json=json, headers=self.headers, cookies=self.cookies, verify=False)`

While far from being optimal, it might be worth considering to apply this change to the repo.
I guess, considering the context in which this is being used, it shouldn't be a a major security threat.

bnkai:
It is probably `safe` enough to override the verify by default (considering the use case).
Thanks for the report will make a PR when possible

scruffynerf:
Solved by #188

PR #188, opened 2023-10-20
[Set Scene Cover] update to use stashapp-tools

stg-annon:
issue #110 addressed within stashapp-tools  `verify_ssl` flag


scruffynerf (approved):
Awesome.  Thanks for a quick turnaround on that.

PR #189, opened 2023-10-20
[phashDuplicateTagger] Updates

stg-annon:
+update how codec's are compared, each codec has a priority level lower value == higher priority, equal priority will be skipped during compare
+add combined file size to dupe group so title sort will show which dupes take up the most space
+add reason tags, reason tags are applied to remove scenes and show why they were not chosen 
+add generate task to generate phashes for all stash scenes
+fixes for [deprecated fields](https://github.com/stashapp/CommunityScripts/pull/189/commits/05b15a7cacf3f635851f0ee2280682ba025796b5)


scruffynerf:
Please don't remove the keep/remove tags.  I use those and due to the way you can only have one filter of each type, sometimes I use searches for other tags and use titles and sometimes I use a search for tags and use a different criteria for titles

stg-annon:
> Please don't remove the keep/remove tags. I use those and due to the way you can only have one filter of each type, sometimes I use searches for other tags and use titles and sometimes I use a search for tags and use a different criteria for titles.
> 
> At least make it optional.

tags are back

scruffynerf (approved):
I'd prefer a way for people to add that doesn't involving editing key files (ideally an imported file not in the repo, so updates don't blow their custom stuff away.

Otherwise, looks great!

PR #190, opened 2023-10-22
updating readme to add the start of v24 upgrade tracking

scruffynerf:
first attempt at prepping for v24 changes - just updating the readme.


scruffynerf:
I do need to update the list of plugins, it's not complete, but I figured I'd just add the new column and remove sqlite renamer from the listing for now.  I'll add the others, as I clean up.

scruffynerf:
merging so we warn people it's broken against develop...

PR #191, opened 2023-10-24
Add CropperJS and Scene Cropper as JS plugins

feederbox826:
original userscript code by @scruffynerf


scruffynerf (approved):
All looks good.  Nice job.

PR #193, opened 2023-11-04
[filenameParser] Add zip format to blocklist

alucicrazy:
This is useful for the galleries that are compacted in a zip file


scruffynerf (approved):
simple one line change.  makes sense.

PR #212, opened 2024-01-03
Handle reverse proxy URL prefix in UserscriptLibrary

GenericMale:
This should make userscripts work on [reverse proxies with a URL prefix](https://docs.stashapp.cc/networking/reverse-proxy/#using-a-url-prefix).
I think this is also how stash core figures out the baseURL ([createClient.ts](https://github.com/stashapp/stash/blob/bd7d4ac7ff857ade8b44e674d87abd3d48f98933/ui/v2.5/src/core/createClient.ts#L114)).


GenericMale:
same as #231

PR #217, opened 2024-01-11
Adding Movie sync support and extra urls from timestamp.trade

Tweeticoats:
If someone has linked a scene to a movie and submitted the data there can be movie information in timestamp.trade. If enabled a movie will be created and the scene number linked on save.

If the extra urls have been submitted to timestamp.trade add them to the scene. I am hoping to crowd source links to funscripts for example.


Maista6969:
Tested this and it works great, looking forward to the addition of more movies! 👍

PR #218, opened 2024-01-11
Adding miscTags plugin for VR

Tweeticoats:
This plugin parses file names for VR useful for stash-vr-companion.

File names in vr have file names like: povr-originals-work-to-liv-or-liv-to-work-180_180x180_3dh_LR.mp4
180x180 indicates 180° projection.
LR indicates it is left right side by side.

For stash-vr-companion add the export_deovr, DOME, SBS tags if you have the setting enabled.
Add the 180°, Virtual Reality or what you have configured in your settings as the vr tag if you have the setting enabled.
I also want to automatically tag 2d scenes based on the studio so set the local studio id in stash ie studio 4 in your instance is Girls Out West so add export_deovr, FLAT to all scenes in this studio.

### DIFF
--- a/plugins/miscTags/miscTags.py
+++ b/plugins/miscTags/miscTags.py
@@ -93,11 +93,15 @@ def processVRTags(scene,tags):
                 found=True
     if found:
         if 'vrTag' in stash.get_configuration()['ui']:
-            vr_tag=stash.get_configuration()['ui']['vrTag']
-            if vr_tag:
-                tags.append(vr_tag)
-            else:
-                tags.append('VR')
+            if 'vrTag' in stash.get_configuration()['ui']:
+                vr_tag=stash.get_configuration()['ui']['vrTag']
+                if vr_tag:
+                    tags.append(vr_tag)
+                else:
+                    tags.append('VR')
+        else:
+            tags.append('VR')
+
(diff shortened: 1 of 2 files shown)


plugins/miscTags/miscTags.py
```diff
@@ -0,0 +1,154 @@
...
+if 'misc-tags' in config:
+    settings.update(config['misc-tags'])
+log.info('config: %s ' % (settings,))
+
+if 'mode' in json_input['args']:
+    PLUGIN_ARGS = json_input['args']["mode"]
+    if 'processScenes' in PLUGIN_ARGS:
+        processScenes()
+    elif 'test' in PLUGIN_ARGS:
+        log.debug(stash.get_configuration()['ui']['vrTag'])
```
Maista6969, line 146: This will crash the user hasn't configured their VR tag, maybe print a nice message instead? Probably not necessary, just a nitpick 😅

Maista6969 (commented):
Looks good to me, but we might want to update those descriptions

plugins/miscTags/miscTags.py
Tweeticoats: I'e updated the code

PR #221, opened 2024-01-16
improve page listeners

Tetrax-10:
## Improve page listeners

Should I set `detectReRenders` as true by default? Because stash technically re-renders the page like when you switch pages.

Formatted the best I can do, using prettier would have saved some time 😅.

https://github.com/stashapp/CommunityScripts/assets/75513645/16d4c406-216e-48ab-8f78-d7954f78e659


feederbox826:
MutationObservers are dangerous and can quickly balloon into memory consumption if you're not careful. From what i've tried, the easiest way to catch navigation is mutationobserver on <head><title> (react hook would be best) and adding eventListernes to the buttons instead of trying to MutationObserver them

Tetrax-10:
Now the the tab pages (non url changes) are handled by click event listeners instead of mutation observers. Only 4 element's changes are detected by mutation observers eg: performer, studio, tag and home edit pages. Additionally all mutation observer will auto disconnect within 10 seconds instead of waiting for elements for infinite time. They also disconnect when there is a page change and element death, Which means there will be no useless leftover mutation observers. Mutation observers only activate on particular pages on particular elements.

Tetrax-10:
For now I will make this as a draft because am trying to implement tagger page listening.

Tetrax-10:
*Re-worte the whole page listener part*

## Features:

1. **Add/Remove custom page event**: We cannot hard-code all page events in the stashUSL. Therefore, let developers add the events they need.
```js
// to add a custom pageListener
stash.addPageListener({event:"custom:page:any:zoom", regex: /.*z=/})

// to listen to the custom event
stash.addEventListener("custom:page:any:zoom", (e) => console.log("Page Zoomed", e.detail))

// to remove a custom pageListener if needed
stash.removePageListener("custom:page:any:zoom")

// to stop listening for page changes
stash.stopPageListener()
```

2. **Added Tagger and View page** support (grid, list, wall)
3. **Detect UI re-renders**. Detects if `.main` element is re-rendered. eg: When you are in scenes page and clicking the scenes nav tab the url wont change but the elements are re-rendered, So with this you can listen and alter the elements inside the `.main` node.
4. **Fixed last event state bug** where `this.lastPathStr` and other variables return the current state to `gmmain()`

Note: I didnt test markers and movies page as I dont have one.

Tetrax-10:
Fixed `stash:page:gallery:images` and `stash:page:gallery:add` and did some touchup.

# Ready to merge !

PR #227, opened 2024-01-20
Add initial version of visage as a plugin

cc1234475:
I converted visage to be a plugin.


becks0815:
Is there a documentation out there explaining how I can use it?

cc1234475:
The plugin is newer, but the video on https://github.com/cc1234475/visage hopeful will be a guide.

becks0815:
Thx, I will look into it, but after searching the Discord server I finally found some comments which also helped. Just for the records: download the plugin within stash and enable it, then install the tampermonkex script (click on the userscript link: https://github.com/cc1234475/visage ), and then you get a new icon in the scene detail view where you can launch the process.

Side question: wouldn't it be easier to create a zip file with faces which can be distributed, so people can build their own face recognition db, or share a list with names so we know which person is already in the db?

cc1234475:
Hi, You shouldn't need the tampermonkey script anymore, this plugin should fully supersede it. I've completely remove tampermonkey from my setup now. The plugin does require a other plugin `Stash Userscript Library`, it should install it as a dependency but it's worth double checking.

There are a few reasons as to why it's setup the way it is.
- Technical limitations: stash plugins are not able to run a server side component and have frontend/backend part. stash has tasks, but visage takes CPU/GPU and takes time to start up (10 secs or so), so running a query through that would proved a very slower user experience.
-  Simple setup: technical limitation aside, I wanted the experience to be simple for others. Not have to worry about python, cuda, pytorch or some of the dependencies the back end needs to deal with, not to mention operating systems.
- Data Freshness: I currently index Stashdb regularly to create the searchable index, As I don't own Stashdb, it's not my place to share it's data more widely. visage provides the basics, like name, country and stashID.
- Making your own database: You most certainly can make your own database, all my code is available https://github.com/cc1234475/visage-ml.
They will not integrate into stash as the visage plugins does as the power of it is leveraging the great work of all the contributors of Stashdb.

cc1234475:
Also feel free to reach to me on discord if you have further questions.

PR #228, opened 2024-01-20
Add initial stashAI plugin

cc1234475:
This is a combination of the stash tag and stash markers userscripts.

The use the same backend API.


Tetrax-10:
This is a compiled version. Do you still have the original code?

cc1234475:
i've not used react, I'll commit the source to a repo.
If someone is open to helping to convert it we can take a look.

PR #232, opened 2024-01-22
Adding a new plugin to automatically download performer images from stashdb

Tweeticoats:
Adding a new plugin to automatically download performer images from stashdb or any other stash-box instance.
Add the tag [Stashbox Performer Gallery] to a performer to start the background job. The images will be downloaded and added to a gallery for the database.

You can then go into the gallery and add the tag [Set Profile Image] to an image with exactly 1 performer to set that image as the profile image.

### DIFF
--- a/plugins/stashdb-performer-gallery/requirements.txt
+++ b/plugins/stashdb-performer-gallery/requirements.txt
@@ -1,0 +1,2 @@
+requests
+stashapp-tools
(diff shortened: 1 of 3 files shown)


github-ph-52obr:
I'm able to use the [Stashbox Performer Gallery] to download the performer images from stashdb. Can the [Set Profile Image] tag only be applied to those images? When I use the tag on another scanned image, the profile image is just a `?`. Do I need to set the download directory to wherever I have all the other performer files? If so, how does the plug in know where to save the stashdb images?

PR #244, opened 2024-02-08
integrate stash-plugin-builder

Tetrax-10:

PR #246, opened 2024-02-10
Make builds deterministic and reproducible

7dJx1qP:
The current `build_site.sh` script generates a new hash for every deployment even if the plugin hasn't changed. The reason the hashes change is due to file timestamps changing when the build checks out the repo and file order when creating the zip.

To address this, I've updated the script so that it sets each file timestamp to the updated date that's already being used and sorts the files that it zips. I also added sorting of the plugins and themes in the index.yml

### DIFF
--- a/build_site.sh
+++ b/build_site.sh
@@ -37,7 +37,8 @@ buildPlugin()
     zipfile=$(realpath "$outdir/$plugin_id.zip")
     
     pushd "$dir" > /dev/null
-    zip -r "$zipfile" . > /dev/null
+    find . -type f -exec touch -d "$updated" {} +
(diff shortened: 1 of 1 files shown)


Tetrax-10:
After this PR #244 this repo no more uses the `build_site.sh` to build the plugins instead a nodejs based `build.js`, so we can build normal plugins and stash-plugin-builder plugins, you can implement this feature [here](https://github.com/Tetrax-10/stash-plugin-builder/blob/main/packages/create-stash-plugin/template/gh-build-workflow/build.js#L132) as `build.js`also uses zip sh under the hood.

PR #248, opened 2024-02-12
[plugin][phashDuplicateTagger] Add ability to compare by created_at

Claypool2112:
This PR adds ability to compare the "created_at" time between files instead of just the "mod_time".

This PR does break some backward compatibility, as "compare_age" is now a bit too broad, so it was renamed to "compare_age_mod_time".  

We could leave "compare_age" alone and simply add "compare_age_created_at" if that is preferable to avoid breakage.

### DIFF
--- a/plugins/phashDuplicateTagger/config.py
+++ b/plugins/phashDuplicateTagger/config.py
@@ -1,7 +1,7 @@
 import stashapi.log as log
 from stashapi.tools import human_bytes, human_bits
 
-PRIORITY = ["bitrate_per_pixel", "resolution", "bitrate", "encoding", "size", "age_mod_time"]
+PRIORITY = ["bitrate_per_pixel", "resolution", "bitrate", "encoding", "size", "age"]
 CODEC_PRIORITY = {
     "AV1": 0,
     "H265": 1,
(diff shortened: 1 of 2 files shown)


plugins/phashDuplicateTagger/config.py
```diff
@@ -121,10 +121,26 @@ def compare_age(self, other):
...
+    if self.created_at == other.created_at:
+        return
+    if self.created_at < other.created_at:
+        better, worse = self, other
+    else:
+        worse, better = self, other
+    worse.remove_reason = "age_created_at"
+    return (
+        better,
+        f"Choose Oldest by Create Time: Δ:{worse.mod_time-better.mod_time} | {better.id} older than {worse.id}",
```
stg-annon, line 143: you'll want to use `created_at` here instead of `mod_time`

stg-annon:
I think the intuitive understanding of "age" for most people is actually your solution so I would just replace age with your method and either remove the current one or leave it around for those that want to use it

not sure why I picked mod_time, must have been a request, it being at the bottom of the list means that I never really used it much myself

stg-annon (approved):

Maista6969:
Closing this as #249 seems to have incorporated the same fix 🙂

PR #249, opened 2024-02-14
[phashDuplicateTagger] Bugfix + Add Compare to Scene paths + Untrack config.py

stg-annon:


stg-annon:
I've incorporated @Claypool2112's #248 changes into this PR as it was easy enough to do while making some changes to the config file so it is untracked by git and the plugin manager

PR #256, opened 2024-02-19
Additional Files Deleter

elkorol:
This is a plugin to replace the Stash-Split-Duplicate-Merged-Files. Which if this pr is merged the Split Duplicate files plugin reference will also need deleted from [https://docs.stashapp.cc](https://docs.stashapp.cc/add-ons/plugins/#stash-split-duplicate-merged-files)

Rather than splitting anymore. One reason if that the old plugin could not work on images as you can't take an image file id and create a new image with it in Graphql. And since most image/scene objects that contain multiple files - these files in the case of scenes have identical Phashes, so are safe to just delete (With exception to manually merged scenes.). In the case of images the multiple files are identical checksums, so are relatively safe to delete and manual merging of image objects is not possible. 

The plugin offers the generation of an ignore tag, to bypass objects with multiple files you don't want deleted.

Apart from the Create Tag task. It offers 4 other tasks. 2 each for scenes and images.

The first kind is just to scan for objects with additional files and delete them
The second kind is to take the paths of the files to be deleted from an object and prefix them with "File: " (For easier searching later) and append them to any current urls if any and update the object.

These tasks are in the case if a user finds that extra files of an object that their paths may contain useful unretrievable metadata if just deleted apart from the primary file. So just to record these in the urls field, since there is no other field adequate for such a thing. 

I know that's not what the urls field is designed for, but is useful in this case.

Thanks


elkorol:
I had to run the formatter a few times even though I ran prettier. It formatted it. But Github Actions didn't like how some debug statements were multi line

DogmaDragon:
@elkorol can you verify this works with the latest Stash release?

DogmaDragon:
[Confirmed as working](https://discourse.stashapp.cc/t/how-to-easily-batch-delete-merged-duplicate-files/891/5?u=dogmadragon).

PR #258, opened 2024-02-21
Added Plex Better Styles CSS

elkorol:
Added Plex Better Styles CSS Theme


Tetrax-10:
Check out #259.

elkorol:
Ah that's even better. Cool man. What about the various other CSS snippets should they be left hardcoded since they aren't that big or should they have their own plugins?

Tetrax-10:
Currently I have managed to dynamically required css files when the option is enabled.

Tetrax-10:
#259 includes Plex Better Styles theme

PR #259, opened 2024-02-21
[plugin - themeSwitch] rm hardcoded css

Tetrax-10:
Currently, the CommunityScripts repository has become bloated due to duplication of themes caused by this plugin. This PR addresses this issue by installing the theme instead of appending hardcoded CSS to the DOM. This way, there is no need to check for updates and copy them to this plugin.


Tetrax-10:
@elkorol d3968b4 : Now css are imported dynamically to reduce plugin size and its much more efficient.

PR #279, opened 2024-03-16
VideoScrollWheel: Add new option for volume change

Tscherno:
allowVolumeChange to let the user decide if volume changing via mouse wheel is enabled. Blocks accidental volume changes.


DogmaDragon:
Thank you for contribution. Tested.

PR #282, opened 2024-04-01
Updates to timestamp.trade and stashdb-performer-gallery

Tweeticoats:
Updating timestamp.trade and stash-performer-gallery for better background processing.

For timestamp.trade scene updates trigger a background task to do the processing so the scene update is not blocked.
A timestamp.trade url can be added to a scene and if this is present it will look up scene metadata for updates.
This allows you to re-process scenes for movie and gallery information after the initial task.

stashdb-performer-gallery has been updated to process in background jobs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_style = space
-indent_size = 4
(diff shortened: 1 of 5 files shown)


Tweeticoats:
This plugin should be ready to merge

PR #307, opened 2024-05-24
use relative path for assets

feederbox826:
- reverse proxy basepath agnostic
- plugin name agnostic

This sets assets to already be at `/plugin/{ID}/`

> Relative URLs, if used, are relative to the URL of the stylesheet (not to the URL of the web page).

https://developer.mozilla.org/en-US/docs/Web/CSS/url


feederbox826:
Ah. Prettier doesn't handle relative paths.

PR #311, opened 2024-05-25
Update markerTagToScene for new Javascript runtime

Maista6969:
Not sure when this happened but the input to Javascript plugins appears to be PascalCase instead of snake_case now so this plugin wasn't working

I also changed it so that it adds not just the primary tag from the marker but also any additional tags, if present

### DIFF
--- a/plugins/markerTagToScene/markerTagToScene.js
+++ b/plugins/markerTagToScene/markerTagToScene.js
@@ -6,9 +6,9 @@ function ok() {
 
 function main() {
   var hookContext = input.Args.hookContext;
-  var opInput = hookContext.input;
-  var primaryTagID = opInput.primary_tag_id;
-  var sceneID = opInput.scene_id;
+  var opInput = hookContext.Input;
+  var primaryTagID = opInput.PrimaryTagID;
+  var sceneID = opInput.SceneID;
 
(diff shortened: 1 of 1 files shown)


p0rnb0x:
hello,
plugin broken now for me. is new JavaScript runtime coming in 0.26.0?
on 0.25.1 this commit breaks plugin for me. getting a lot of errors like this:

```
/config/plugins/community/markerTagToScene/markerTagToScene.js: Line 26:18 Unexpected token . (and 32 more errors)
```

Maista6969:
I apologize for the inconvenience: the new runtime is available in `latest_develop` and will be in `v0.26.0`
https://github.com/stashapp/stash/compare/v0.25.1...latest_develop

In the mean time you can revert your plugin file to the [previous version](https://raw.githubusercontent.com/stashapp/CommunityScripts/228c294ffa30b191c50ac147af319c8990a9395e/plugins/markerTagToScene/markerTagToScene.js)

p0rnb0x:
thank you friend!

strong-path:
Thanks so much for the update @Maista6969, this plugin is great. Just a quick note because it always throws this error after running successfully: `job.progress: input: job.progress cannot marshal infinite no NaN float values\ job.progress: null`. Is this just me or is this normal?

PR #312, opened 2024-05-26
No cropper js

feederbox826:
Import CropperJS by URL instead of with module


DogmaDragon:
What are the advantages of this? From my point of view, it has a disadvantage of not being able to work offline anymore.

feederbox826:
Advantages
- We no longer have to ship the entire codebase unminified
- We won't have potential version conflicts with 7dJx1qP CropperJS

Disadvantages
- No longer fully offline

Notable points
- 7dJx1qP has already switched to URL
- CropperJS has a v2 overhaul on the way and it will just wreak havoc if both are enabled

The case for fully-offline vs CDN'd modules is... tricky.... I would rather uniformly rely on CDNs rather than mix the two and discover version conflicts

DogmaDragon:
I agree that avoiding conflicts should be the priority. It's sad to lose offline functionality, but it's inevitable I guess.

PR #314, opened 2024-05-30
Add AITagger plugin to community repo

skier233:
This adds a new plugin called AITagger which can tag images and scenes in batch using Skier's AI models. See more info here:
https://www.patreon.com/posts/ai-tagging-of-104503565
https://github.com/skier233/nsfw_ai_model_server


plugins/AITagger/ai_tagger.yml
```diff
@@ -0,0 +1,17 @@
+name: AI Tagger
+description: Tag videos and Images with Locally hosted AI using Skier's Patreon AI models
+version: 1.0
+url: https://github.com/stashapp/CommunityScripts/tree/main/plugins/AITagger
+exec:
+  - python
+  - plugins/AITagger/ai_tagger.py
```
feederbox826, line 7: if you reference the file instead of the base dir, you can avoid the chdir

plugins/AITagger/ai_tagger.py
```diff
@@ -0,0 +1,345 @@
+import os
+import sys
+import json
+import subprocess
+import csv
+from typing import Any
+
+# plugins don't start in the right directory, let's switch to the local directory
+os.chdir(os.path.dirname(os.path.realpath(__file__)))
```
feederbox826, line 9: chdir not necessary if you do relative path in .yml

plugins/AITagger/ai_tagger.py
```diff
@@ -0,0 +1,345 @@
...
+
+# ----------------- Setup -----------------
+
+def install(package):
+    subprocess.check_call([sys.executable, "-m", "pip", "install", package])
```
feederbox826, line 14: will likely raise errors on docker container without --break-system-packages and bypasses venv fyi

plugins/AITagger/config.py
```diff
@@ -0,0 +1,9 @@
+API_BASE_URL = 'http://localhost:8000'
```
feederbox826, line 1: Consider putting this in a UI setting if it's going to be a commonly changed value - plugin updates are not merged on the client

feederbox826 (commented):
A few quirks and untested without licence

plugins/AITagger/config.py
skier233: This will never get changed on my end and is only there for the user if they wanted to change it

plugins/AITagger/ai_tagger.py
skier233: is there currently a better best-practice way to auto-install dependencies?

plugins/AITagger/ai_tagger.yml
skier233: I'm not exactly sure what you mean. Can you give an example?

plugins/AITagger/ai_tagger.py
skier233: See above comment

plugins/AITagger/ai_tagger.yml
feederbox826: https://github.com/stashapp/CommunityScripts/blob/42c7da20754477c5e7b5abff84e502ab707b1c06/plugins/phashDuplicateTagger/phashDuplicateTagger.yml#L5-L7

plugins/AITagger/ai_tagger.py
feederbox826: it is a mess righ tnow, probably not a better way but including a requirements.txt I believe will trigger py_common and stash-s6

plugins/AITagger/ai_tagger.yml
skier233: I tested that approach and with it and removing chdir it broke accessing the local path to the csv. Or are you saying that I can also use that {pluginDir} in the python file as well instead of a relative path?

plugins/AITagger/ai_tagger.py
skier233: I'll add a requirements.txt and surround this install in a try catch which gives a more user-friendly message if they try to install on docker or an environment where this wont work with a message detailing how to manually install the dependencies if all else failed.

PR #317, opened 2024-06-01
rewrite marker grouping algo, use pluginDir

feederbox826:
requesting review from @skier233

- bump MaxGap to 2 since model seems to generate at intervals of 2

## rewrote marker tagging algorithm
When I was trying out the plugin, I could never get any tags to appear since for some reason they would only have a duration of 0s

Instead of walking through it and searching backwards per marker, this instead searches forwards.

This new sorting algorithm walks through all of the marked timestamps, per tag and continues until the 2s gap is exceeded or it reaches the end. If the gap is exceeded, that marker is closed off and the next one started unless it exceeds the number of timestamps left. Some variables were also renamed in order to more clearly reflect their purpose

### DIFF
--- a/plugins/AITagger/ai_tagger.py
+++ b/plugins/AITagger/ai_tagger.py
@@ -6,8 +6,6 @@ import csv
 from typing import Any
 
 # ----------------- Setup -----------------
-os.chdir(os.path.dirname(os.path.realpath(__file__)))
-
 def install(package):
     try:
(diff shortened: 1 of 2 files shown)


skier233:
Incorporated these changes and a few other fixes into https://github.com/stashapp/CommunityScripts/pull/318

DogmaDragon:
Closed in favor of https://github.com/stashapp/CommunityScripts/pull/318.

PR #326, opened 2024-06-09
Update defaultDataForPath.js object.ID has changed to object.id

ThirstilyVague:
I think there are a few more fixes to make, but this is one of the first I've found so far.


ThirstilyVague:
Sorry I don't think this is going to fix it.

PR #343, opened 2024-06-21
Tag scenes with performer tags plugin

Dankonite:
This is my plugin for taking performer tags (race, breast size, tattoos etc. etc.) and applying them to scenes they appear in.


Dankonite:
Idk if there are any things im missing for a proper submission, I cant tell how manifest files work (whether I make them or they get made automatically?)

DogmaDragon:
Made automatically based on .yml file.

PR #349, opened 2024-06-28
[hotcards] add ?. fallbacks and clean regex, fix dependency chain

feederbox826:
- use ?. to not fail when loading from `{}`
- use Object.values instead of Object.entries
- clean up regex with repeating patterns
- use pathname to pre-strip parameters
- move css to bottom, to ensure depencies load first

### DIFF
--- a/plugins/hotCards/hotCards.js
+++ b/plugins/hotCards/hotCards.js
@@ -2,43 +2,42 @@
   "use strict";
 
   const userSettings = await csLib.getConfiguration("hotCards", {});
-  const TAG_ID = userSettings.tagId;
-  const RATING_THRESHOLD = userSettings.threshold;
+  const TAG_ID = userSettings?.tagId;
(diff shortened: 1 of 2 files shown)


plugins/hotCards/hotCards.js
```diff
@@ -159,7 +158,7 @@
    */
   function handleScenesHotCards() {
     const pattern =
-      /^\/(scenes$|performers\/\d+\/scenes$|studios\/\d+\/scenes$|tags\/\d+\/scenes$|movies\/\d+$|galleries\/\d+$)/;
+      /^\/(scenes|(performers|studios|tags|movies)\/\d+\/scenes|(movies|galleries)\/\d+)$)/;
```
HandyRandyx, line 161: I think you left an extra ')' right at the end. I just updated the plugin and I'm getting "SyntaxError: unmatched ) in regular expression". Thank you.

HandyRandyx: #350

PR #355, opened 2024-07-04
[Hot Cards] Customizable Hot Cards

HandyRandyx:
Key features added:
- Customizable Hot Cards.
- Flexibility to choose tag-based or rating-based for each card type with different values if desired.
- Two more style presets in addition to the default.
- Support for 0-5 explicit Rating Threshold alongside 6-100 range for tenths.


feederbox826:
LGTM at a glance

PR #375, opened 2024-07-28
Added plugin FileMonitor and updated RenameFile plugin

David-Maisonave:
### Plugin FileMonitor.
I found the **stash-watcher** script after I started implementing FileMonitor, but stash-watcher has some issues.
#### stash-watcher Issues
- It doesn't work as-is. I had to make some changes in the code to get it working.
- After fixing it, I notice that when a change occurs in the file system, **stash-watcher** performs a full scan on everything in the library, instead of just scanning the path (directory) with the change.
- **stash-watcher** does **NOT** fetch the stash library content, and instead user has to populate a *.toml file with desired paths. If the library content changes, the user also has to change the *.toml file.
- Can not start the script from the Stash GUI. Not user friendly since user has to be familiar with launching python scripts.

#### FileMonitor
FileMonitor is similar to the stash-watcher script, put it works like a Plugin.
- The user can start it from the Stash UI in [Plugins Task]. And the user can kill it from the [Task Queue]
- **FileMonitor** automatically gets all the Stash library paths, and monitors those paths.
- When a change occurs, it calls **stash.run_plugin_task** with the directory of the changed file, so only the effected path is scanned.
- FileMonitor runs out of the box without the user having to make any setting/configuration changes.
- FileMonitor can also optionally run as a standalone script.
  - To start monitoring, call the script and pass any argument.
    - python filemonitor.py **start**
  - To stop **FileMonitor**, pass argument **stop**.
    - python filemonitor.py **stop**
    - After running above command line, **FileMonitor** will stop after the next file change occurs.
    - The stop command works to stop the standalone job and the Stash plugin task job.
  
### Update to RenameFile plugin.
- Added fields galleries, resolution, and width.
- Fixed bug associated with studio.
- Added logic to limit the log file size.
- Added logic to only get [Change success] logging when no errors occurs.
- Change default fields to include studio.
- Added postfix styles, which was mainly needed to properly format resolution field, but can be used for the other fields.
- Consolidated [Include Existing Key Field] options into one option.
- Cleaned up code and updated version


DogmaDragon:
Please remove manifest files. They will be generated by the build action. Also .gitignore seems to be related to debug/development, not relevant for regular users.

David-Maisonave:
> Please remove manifest .... .gitignore .....

Removed.

PR #393, opened 2024-08-10
New issue templates

David-Maisonave:
I recommend using YML files over MD templates because they have required fields which forces users to populate them before submitting issues, and the interface is more user friendly, and the issue titles use icons which can help differentiate between different types of issues.

To see how they'll look, check out the following link:
[Issues](https://github.com/David-Maisonave/Axter-Stash/issues/new/choose)

I included more than what you probably want, so please feel free to kill the YML files you don't want.


David-Maisonave:
Closing this pull request so as to allow plugin changes to go through first.

PR #416, opened 2024-08-24
[markerDeleteButton] Delete buttons on Scene page next to Edit button

MinasukiHikimuna:
This is especially useful when reviewing markers created by Skier's AI model.

Known issues: If a new marker is created or an existing marker edited, Delete buttons are no longer visible when returning to the page. csLib.PathElementListener only activates when the user navigates to the page after which the specific element must appear. When moving from Create or Edit forms back to the list of markers there is no navigation event and thus Delete buttons will not be re-created.

### DIFF
--- a/plugins/markerDeleteButton/markerDeleteButton.js
+++ b/plugins/markerDeleteButton/markerDeleteButton.js
@@ -2,7 +2,7 @@
   const markerDeleteButton =
     '<button class="marker-delete-button btn btn-danger">Delete</button>';
 
-  async function setupMarkerDeleteButton() {
+  async function setupMarkerDeleteButtonForMarkersWall() {
     document
       .querySelectorAll("div.wall-item-container")
       .forEach(function (node) {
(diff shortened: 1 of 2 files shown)


WeedLordVegeta420:
@MinasukiHikimuna I honestly hadn't considered using this from the Scene page when I made this plugin, but it looks like a good addition. Thanks for the update.

sbaak4:
for the scene page

using pop would make this support -'s in the tag itself  
tag:`foo - bar` becomes `<div>foo - bar - 28:18</div>`
```diff
- const markerTime = node.querySelector(".wall-item-text div").innerText.split(" - ")[1];
+ const markerTime = node.querySelector(".wall-item-text div").innerText.split(" - ").pop();
```

the `.btn.btn-link` is the markers title if it has a title, if no title is set then its the tag. since we are making the markerMap with the tag it fails with titles that dont match the tag.
the h3 above it is always the primary tag of the marker/s in that card. 
```diff
- const markerTag = node.querySelector(".btn.btn-link").innerText;
+ const markerTag = node.closest('.primary-card').querySelector('h3').innerText;
```

just leaving this somwhere incase its helpful. worked for me but i didnt set up a clean container. so ya who knows.

PR #426, opened 2024-09-04
Updates to Timestamp Trade

vamiceDev:
Added some new options to Timestamp Trade
Option to allow the user to add a tag to all generated timestamps
Option to allow the user to add a prefix to the name of all generated timestamps
Option to run on scenes that already have timestamps, with code to identify and ignores duplicates.


vamiceDev:
Wait, which conflicting usernames?  My apologies if I missed something.

PR #428, opened 2024-09-05
Updates to Timestamp Trade, Attempt 2

vamiceDev:
I'm not that great at Git, a skill that's on my list to learn. So I just made a new repo with the same name and a single commit.

@feederbox826 Thanks for the heads-up.

Updates TimestampTrade to allow the user to add a tag/update the title of created timestamps.

### DIFF
--- a/plugins/timestampTrade/timestampTrade.py
+++ b/plugins/timestampTrade/timestampTrade.py
@@ -72,33 +72,7 @@ def processSceneTimestamTrade(s):
                                 getTag("[TsTrade]")
                                 marker["tags"] = ["[TsTrade]"]
                             if len(marker["primary_tag"]) > 0:
-                                if settings["runOnScenesWithMarkers"]:
-                                    found = False
-                                    cur_marker_string = ";".join(
-                                        [
-                                            str(marker["title"]),
-                                            str(int(marker["seconds"])),
-                                            str(getTag(marker["primary_tag"])),
-                                        ]
-                                    )
-
-                                    for m in s["scene_markers"]:
-                                        scene_marker_string = ";".join(
-                                            [
-                                                str(m["title"]),
-                                                str(m["seconds"]),
-                                                str(m["primary_tag"]["id"]),
-                                            ]
-                                        )
-                                        if cur_marker_string == scene_marker_string:
-                                            found = True
-                                            break
-                                    if found:
-                                        log.debug("Duplicate marker, ignoring")
-                                    else:
-                                        markers.append(marker)
-                                else:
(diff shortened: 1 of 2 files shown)


feederbox826:
your commits were not under `vamiceDev`, figured I would give you a heads up. [tips for splitting identities](https://github.com/feederbox826/notes/blob/main/seperate-identity.md)

also cc @Tweeticoats

vamiceDev:
Yes, I saw right after I submitted, I appreciate it.

Tweeticoats:
runOnScenesWithMarkers is not a flag that I want because it's already covered by the setting  createMarkers.
The plugin should run on the scenes as there are other functions that the user may want so the user could disable creating markers but still have the plugin to create movies.

I'm not against adding a tag and adding a prefix to markers.
For consistency I would use [Timestamp]  or similar as the plugin uses [Timestamp: Skip Submit], [Timestamp: Skip Sync], [Timestamp: Auto Gallery] so it would be more consistent.

I'm working on my own pull request so I'll add the changes and submit that instead if that is ok.

vamiceDev:
> runOnScenesWithMarkers is not a flag that I want because it's already covered by the setting createMarkers.

I don't see how.  The query run when "processScene" == PLUGIN_ARGS uses "has_markers": "false".  Also, there's a check in the processSceneTimestamTrade function that checks if len(s["scene_markers"]) == 0.

The use case I'm targeting here is a user who made a custom marker for a scene and now wants to add extra markers from timestamp.trade.  As it's currently written, that scene will be skipped.

> For consistency I would use [Timestamp] or similar as the plugin

Sure that's fine.

> I'm working on my own pull request so I'll add the changes and submit that instead if that is ok.

Yeah, that's fine, but I would like to see the functionality from runOnScenesWithMarkers  added.

Tweeticoats:
I have multiple tasks:
 - Sync:   just run on scenes without markers
 - Re-process Scene:   look for scenes with a timestamp.trade/scene/uuid  url to re-process scenes that have been processed before
 - Re-process All:  run on all scenes with a stashid and this is what I would recommend you run most of the time

If a scene has markers it will skip the scene unless you enable overwriting markers.
I'm happy to have a merge setting if you would find it useful.

vamiceDev:
I see my mistake now, missed the Reprocess All hook, so that's fine.

Suggestion: Calling it **Re**-Process All implies it only runs on previously processed scenes.

Having an option to merge would be preferable, and I'd argue that it should be on by default.

Also, overwriteMarkers is not in the timestampTrade.yml file, so it's not in the Stash UI.

PR #430, opened 2024-09-09
A few more features, funscripts matching, excluding markers with matching strings and changes to marker titles and tags

Tweeticoats:
A few more features, funscripts matching, excluding markers with matching strings and adding a string to new markers.

Indexing funscripts is done with a task creating a database with the funscript path and a md5 of the script. Submitting funscripts is done by the submit tasks. On query a list of funscript hashes are returned by the api and if thre is a matching script it will copy the file to the correct name for stash to match it.

Added an option to exclude markers that contain a string.

Added an option to add the [Timestamp] tag to markers the plugin creates.

Added an option to add the string [TsTrade] to markers eg "[TsTrade] marker name".

And bugfixes

### DIFF
--- a/plugins/timestampTrade/timestampTrade.py
+++ b/plugins/timestampTrade/timestampTrade.py
@@ -8,6 +8,10 @@ import time
 import math
 import uuid
 from pathlib import Path
+import sqlite3
+import hashlib
+import shutil
+import re
 
 per_page = 100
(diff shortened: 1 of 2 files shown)


vamiceDev:
@Tweeticoats Thanks!

feederbox826:
Supercedes #428

PR #436, opened 2024-09-24
Fix syntax error in timestampTrade.py

Written2001:
This line leads to an syntax error: "unmatched '['" due to the usage of double quotes for both the f-string and the inner dictionary key.
The proposed solution fixes that.

### DIFF
--- a/plugins/timestampTrade/timestampTrade.py
+++ b/plugins/timestampTrade/timestampTrade.py
@@ -66,7 +66,7 @@ def processSceneTimestamTrade(s):
                                 marker["primary_tag"] = m["name"]
 
                             if settings["addTsTradeTitle"]:
-                                marker["title"] = f"[TsTrade] {m["name"]}"
+                                marker["title"] = f"[TsTrade] {m['name']}"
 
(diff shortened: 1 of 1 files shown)


Maista6969:
Nested quotes are supported in Python 3.12 and above

Written2001:
Not sure if it's the default but the python version in my stash docker container is 3.11.2

feederbox826:
Ah. Nerethos is still on py3.11.2

Tweeticoats:
Another bug that needs to be fixed: the "path" setting  is not configured in the defaults.
The UI has this as "Download parent folder" but it is not set until it is modified so it needs to be added to the settings default  object.
Could you also add the following to  the settings on line 1335
`    "path": "/download_dir/tt-auto",`

Written2001:
> Ah. Nerethos is still on py3.11.2

Yeah the Image is based on debian bookworm which has no prebuild deb packages for python3.12

feederbox826:
superceded by https://github.com/stashapp/CommunityScripts/commit/3e4fb6383e8a405ad5ed2b1d47502c75bd8d42de

PR #450, opened 2024-10-04
[Hot Cards] Rename Movie to Group and enhance homepage loading

HandyRandyx:
Notable:
- Renames `movie` to `group`. Adds new `subgroups` path.
- It is no longer necessary to refresh the page to see the new style if the user updates their plugin settings.
- Removes 3 second delay from homepage. Now uses a mutation observer. Once all cards of a given type have been loaded through all the recommendation rows, the hot cards are added.

Other:
- Improves card border when not using hover animation.
- Reorganize code.

I've been testing it and everything seems to be working fine, but let me know if you find anything odd.


feederbox826:
LGTM, only nitpicks being that some stuff can be a class/css var

PR #458, opened 2024-11-05
[markerTagToScene] reduce logging

p0rnb0x:
This plugins logs lot of JSON and other info. I disable JSON log and convert other one from Info to Debug.

### DIFF
--- a/plugins/markerTagToScene/markerTagToScene.js
+++ b/plugins/markerTagToScene/markerTagToScene.js
@@ -78,7 +78,7 @@ function processMarker(marker, shouldHandleAllTags) {
 }
 
 function main() {
-  log.Info(JSON.stringify(input));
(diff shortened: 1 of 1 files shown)


DogmaDragon:
I don't think having less complete log by default is preferred. If you feel like it's excessive you can modify your plugin locally.

And the change from Info to Debug doesn't make sense in that content. It's just reporting the mode used. Unless I misunderstand something?

p0rnb0x:
> I don't think having less complete log by default is preferred.

neither is dumping everything at `Info` level.

> If you feel like it's excessive you can modify your plugin locally.

already have; feel free to close this PR if you consider the current log statements appropriate at `Info` level.
 
> It's just reporting the mode used.

correct; which obviously is for `Debug`ging

PR #467, opened 2024-11-23
DupFileManager 0.1.9

David-Maisonave:
Added the following features:
- Duplicate File Report which displays duplicates and has buttons to perform individual deletion, copy, rename, metadata copy, flagging, tagging, and more.
- Tools UI option which allows access to the new Duplicate File Report, and additional Tools and Utilities options.
- Advance menu option.
  Screenshot 2024-11-22 145139
  - Delete only specially tagged duplicates in blacklist path.
  - Delete duplicates with specified file path.
  - Delete duplicates with specific string in File name.
  - Delete duplicates with specified file size range.
  - Delete with specified duration range.
  - Delete with resolution range.
  - Delete duplicates having specified tags.
  - Delete duplicates with specified rating.
  - Delete duplicates with any of the above combinations.
- Bottom extended portion of the Advanced Menu screen.
  - Screenshot 2024-11-22 232005


DogmaDragon:
Can you remove the dev files?

David-Maisonave:
Good catch.  Removed.

David-Maisonave:
After running prettier, DupFileManager.js is really hard to read, because it stretches out the jquery code into  1000's of lines of code, instead of leaving it isolated into on line.
Is there someway to get prettier to ignore a line of code?

I know that the right way to do this JS code is to reference the jquery code externally, but Tools UI doesn't allow for any external referencing.

David-Maisonave:
Fixed prettier issues.

PR #470, opened 2024-11-28
[DupFileManager] plugin updated to version 1.0.0

David-Maisonave:
### 0.2.0
- For report, added logic to transfer option settings **[Disable Complete Confirmation]** and **[Disable Delete Confirmation]** when paginating.
- Fixed minor bug in advance_options.html for GQL params.
### 0.2.1
- Added logic to have reports and advanced menu to work with Stash settings requiring a password by adding API-Key as param argument for advance menu, and adding API-Key as variable in reports.
- When **[Advance Tag Menu]** is called from reports, it's given both the GQL URL and the apiKey on the URL param, which allows advance menu to work with non-standard URL's and with API-Key.
### 0.2.2
- Added dropdown menu logic to Advance Menu and reports.
- Added Graylist deletion option to Advance Menu.
- Report option to clear all flags from report.
- Report option to clear all (_DuplicateMarkForDeletion_?) tag from all scenes.
- Report option to delete from Stash DB all scenes with missing files in file system.
- Added popup tag list to report which list all tags associated with scene.
- Added popup performer list to report which list all performers associated with scene.
- Added popup gallery list to report which list all galleries associated with scene.
- Added popup group list to report which list all groups associated with scene.
- After merging tags in report, the report gets updated with the merged scene metadata.
- Added graylist deletion option to [**Advance Duplicate File Deletion Menu**].
- Added pinklist option to Settings->Plugins->Plugins and to [**Advance Duplicate File Deletion Menu**]
  - The pinklist is only used with the [**Advance Duplicate File Deletion Menu**], and it's **NOT** used in the primary process to selected candidates for deletion.
- Advance Menu now works with non-tagged scenes that are in the current report.
### 1.0.0
- Consolidated buttons and links on report into dropdown buttons.
- On report, added dropdown menu options for flags.
- Rename Tools-UI advance duplicate tagged menu to [**Advance Duplicate File Deletion Menu**]
- When [**Advance Duplicate File Deletion Menu**] completes report, gives user prompt to open the report in browser.
- Added performance enhancement for removing (clearing) duplicate tags from all scenes by using SQL call.
- Added option to report to delete files that do not exist by duplicate candidates in report, as well as by tagged files.
- Added logic to disable scene in report if deleted by [**Advance Duplicate File Deletion Menu**]. Note: Requires a refresh.
- Added report option to delete by flags set on the report.
- Fixed bug with report delete scene request.
- In the report, made icon colors for tags, performers, galleries, and groups with different colors if they don't match. In other words, use different color icons if **candidate to delete** doesn't match **duplicate to keep** associated icon data.
  - If data for associated icon are the same, then both icons are black or blue (the default color).
  - If [**duplicate to keep**] is missing data that is in [**candidate to delete**], than [**candidate to delete**] gets a yellow icon.
  - If [**candidate to delete**] is missing data that is in [**duplicate to keep**], than [**duplicate to keep**] gets a pink icon.
- Added option on report to merge all metadata missing in [**Duplicate to Keep**] files.
- Added cookies to report so as to remember user options for Disable Complete Confirmation **[Disable Complete Confirmation]** and **[Disable Delete Confirmation]**.
  - This change was needed because sometimes the browser refuse to open local URL's with params on the URL.
  - Using cookies also allows check options status to stay the same after refresh.
- Added code to [**Advance Duplicate File Deletion Menu**] to delete based on flags.


David-Maisonave:
prettier seems to have a bug in it, because it's not recognizing the matching TR in the HTML code.

Maista6969:
> prettier seems to have a bug in it, because it's not recognizing the matching TR in the HTML code

I think it's because the `</tr>` on line 308 is unintentionally matching the `<tr>` on line 282. Judging by the indentation you either have an extra `</tr>` on the end of line 285 or you meant to add another `<tr>` on line 286

David-Maisonave:
<TR> on 282 matches </TR> on 321
<tr style="border: none;"> on 284 matches </TR> on 308
<TR> on 285 matches another </TR> on the same line 285

I checked, and all the <TR>'s have a matching </TR>.  I don't see an extra.
I also performed a count by doing keyword search for "<TR" which gave me 16, and keyword search for "</TR>", which also gave me 16.
FYI: You can not do a keyword search for "<TR>" because that will miss the TR with a style in it.

I initially thought the TR with the style property was confusing prettier, but even when I converted that to a normal <TR>, prettier still gave the same error.  I also tried lining up the TR via tabbing, and I also tried adding ****, but nothing seems to work to make prettier happy.

Maista6969:
I personally don't think the prettier lints are adding a lot to this repo, but I also tried to drop your HTML into the [W3 HTML Validator](https://validator.w3.org/nu/) and it's lighting up that same `<tr>` element for a different reason 😅

David-Maisonave:
OK,
I'll look into this some more.  There must be something wrong, and it's probably just giving wrong or misleading error.

DogmaDragon:
> <TR> on 282 matches </TR> on 321
> <tr style="border: none;"> on 284 matches </TR> on 308
> <TR> on 285 matches another </TR> on the same line 285

Line 284 <tr> is being closed on line 285
Line 285 <tr> is being closed on line 308.

You can't nest them. You need to create table inside table for that.

David-Maisonave:
As I thought.  It was a matter of the error being misleading. 
I'll revisit the errors listed in [https://validator.w3.org](https://validator.w3.org) on another update.
That's a very useful tool.

PR #507, opened 2025-02-26
Parse All Galleries for date

dynamosTopspin:
- Break out the date parsing to a separate function.
- Expand gallery search to more than single file zip galleries.
- Also scan the gallery's folder path if available.

The final valid parsed date of the files, then the gallery folder will be taken as the gallery's date. This will preserve the previous behavior of single file zip galleries, as there will be only one path to ever consider.

### DIFF
--- a/plugins/DateParser/date_parser.py
+++ b/plugins/DateParser/date_parser.py
@@ -22,15 +22,19 @@ def main():
         find_date_for_galleries()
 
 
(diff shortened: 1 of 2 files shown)


dynamosTopspin:
I found a bug in my change where a parent directory with a valid date string would return that value when a sub directory or file might have a different date.

PR #516, opened 2025-03-19
[hotCards] added silver and bronze default card style

CrookedDuck:
this adds two new style presets (silver and bronze) based on the style of the 'gold' preset. 
for example: r_5/4/3/2_gold/silver/bronze/none____///,50

image

i thought i would be useful for some as it was for me. 
now you can easily have a gold-silver-bronze rating for your hotCards.


DogmaDragon:
Please increase the plugin version in the .yml file.

PR #519, opened 2025-03-30
Fixes 3D banner enabling Profile Cards

cj12312021:
The CSS for the Profile Cards was mixed into the CSS for the 3D banners. This change separates the two, so enabling 3D banners does not enable Profile Cards


DogmaDragon:
Can you bump the version?

PR #520, opened 2025-04-04
add CODEOWNERS

feederbox826:
formalize CODEOWNERS


DogmaDragon:
I would prefer to use https://discourse.stashapp.cc/ for that. And git history will preseve the authors for those that are no longer active community members. On top of that both plugin and userscript configuration files allows declaring the author directly.

PR #521, opened 2025-04-04
WIP Add support for end seconds

WorksDontBreak:
This should In theory work currently testing

### DIFF
--- a/plugins/timestampTrade/timestampTrade.py
+++ b/plugins/timestampTrade/timestampTrade.py
@@ -51,6 +51,7 @@ def processSceneTimestamTrade(s):
                         previous = {"title": "", "seconds": -1}
                         for m in data["markers"]:
 
+
                             marker = {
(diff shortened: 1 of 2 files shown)


WorksDontBreak:
I had to migrate my database to make this work locally, if im not mistaken if you run this from scratch it should just make the database correctly from the start

WorksDontBreak:
The script now supports end_seconds, and it pushes end_seconds but dosent get them when syncing, the endpoint it uses need to be updated to work with this

feederbox826:
cc @Tweeticoats

DogmaDragon (changes requested):
Needs a rebase. 

Is there any need to merge this without the `timestamp.trade` database supporting `end_seconds`? As the task will never produce anything.

DogmaDragon:
Closing as stale.

PR #534, opened 2025-04-21
Remove QxSceneCard in favor of native repo

feederbox826:
@QxxxGit

we discussed this briefly in the discord, If you're ready to disassociate it just let me know


QxxxGit:
Yeah, it can be removed. Thanks for taking care of this.

The plugin can be installed with my source URL `https://qxxxgit.github.io/stash-plugins/index.yml`

PR #539, opened 2025-04-24
[tagScenesWithPerfTags] Add check if key exists

blaspheme-ship-it:
This checks for `inputFields` on `hookContext`. This prevents an error that happened when scanning scenes for the first time.


DogmaDragon:
Please bump the version in the .yml file.

PR #541, opened 2025-04-30
[plugins/miscTags] VRCompanionTags Not running

ozlabrat:
Was never taging based on VRCompanionTags as the setting name is case sensitive.


DogmaDragon:
Can you bump the version in the .yml file?

ozlabrat:
Version has been bumped to 0.5

PR #558, opened 2025-05-17
e621_tagger new features

yokarion:
- added `e621_tag_failed` - this tag appeares of its failed to tag the image (exception or e621 api error)
- marking images as categorized when tagged
- added pagination to prevent OOM Killing


DogmaDragon:
Can you bump the version in the .yml file?

PR #559, opened 2025-05-18
Add forum links in README files for various plugins and themes

DogmaDragon:


feederbox826:
just a little bit late, but discourse can work with just /t/id, https://discourse.stashapp.cc/t/1327 is valid and works

DogmaDragon:
> just a little bit late, but discourse can work with just /t/id, [discourse.stashapp.cc/t/1327](https://discourse.stashapp.cc/t/1327) is valid and works

Yeah, everything in the middle doesn't matter, it's just for SEO purposes. As long as the id is valid it will redirect.

PR #562, opened 2025-05-24
bulkImageScrape plugin now uses generator and paging to iterate images

spaceyuck:
The bulkImageScrape plugin used to request data of all images in the library in one big query at the beginning, which got really problematic on a very large library (like 250K images), leading to stash to first consume all the memory it can get and then die.

The new behavior uses page queries and a generator to process images in batches of 100.

I also added a plugin setting to skip a set number of entries, this was useful for runs that die midway through (for example due to scraper issues, etc).

I have no idea about the version number, so I just bumped it to a .1


NotMyMainUser:
Looks nice, thanks for that 👍 💯

PR #563, opened 2025-05-28
Add PythonDepManager plugin

NotMyMainUser:
Python dependency management system for CommunityScripts plugins (pretty much what py_common does for scrapers)

This plugin provides an easy way to install and manage Python package dependencies in your plugins without manual user interaction.
Don't worry about missing dependencies and wrong or conflicting versions anymore


feederbox826:
Don't like that this even has to be done, but glad we have another one I guess

PR #566, opened 2025-05-29
[StashRandomButton] Add support for scenes, images, performers, studios, tags, groups and galleries

MrDocSabio:
This PR refactors and extends the Random Button plugin to support random navigation for all major entities in StashApp (scenes, images, performers, studios, groups, tags, galleries), both globally and internally. 

- Adds random selection for all entities and internal entity navigation (e.g. random scene in a studio, random image in a gallery).
- Updates README and manifest to reflect the new features.

Tested on v0.27.2+. Please let me know if any changes are required!

### DIFF
--- a/plugins/StashRandomButton/README.md
+++ b/plugins/StashRandomButton/README.md
@@ -1,13 +1,23 @@
 # Stash Random Button Plugin
 
-https://discourse.stashapp.cc/t/randombutton/1809
+[Plugin thread on Discourse](https://discourse.stashapp.cc/t/randombutton/1809)
 
-Adds a "Random" button to the image & scenes page to quickly navigate to a random scene.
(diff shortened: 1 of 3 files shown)


MrDocSabio:
You're correct. At the moment, the random button is not working on object pages (scenes or images).

I initially disabled it because keeping track of filters or context (like coming from a specific tag, studio or gallery) can be tricky. But I understand that users expect the Random button to be available everywhere, even if it just jumps to a completely random scene/image/etc.

However it is possible to implement the Random button so that, when you're on a detail page, clicking Random will simply take you to a completely random item (scene, image, etc) from all available items, regardless of the previous context.

I will try to implement this functionality asap, so you can continue randomizing even when viewing the details of a scene or image.

Thanks for your feedback m8!

PR #569, opened 2025-05-30
Add PlexSync plugin

HarryS:
This is a plugin that is used in conjunction with https://github.com/Darklyter/StashPlexAgent.bundle -- despite this "constraint", it seems relevant to have it here.

The main purpose is that the Stash plugin refreshes metadata on Plex, when the relevant scene is updated in Stashapp. It also optionally cleans titles, and adds tags to scenes based on what stashbox it was sourced from.

Thanks!

### DIFF
--- a/plugins/PlexSync/PlexSync.py
+++ b/plugins/PlexSync/PlexSync.py
@@ -1,0 +1,147 @@
+import stashapi.log as log
+import re
+from stashapi.stashapp import StashInterface
+import sys
+import json
+import unidecode
+import requests
+
+
+def processScene(scene):
+    tags = []
+
+    if settings["skipUnorganized"] and not scene['organized']: # don't process scenes that aren't marked organized.
+        log.info("Not organized, bye!")
+        sys.exit()
+
+
+    if settings["tagStashIDs"]:
+        if not scene["stash_ids"]: # Tagging empty Stash ID
+            tags.append(empty_stashid)
+
+        for stashbox in scene["stash_ids"]: # Add all the stashbox ID tags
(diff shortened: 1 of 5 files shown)


HarryS:
> Seeing as this is supposed to be a plugin coutnerpart of the plex agent, I'm not sure it really fits into communityscripts

(Just adding here that I replied on Discord "it has some more features than that; but should stuff that have external dependencies not be in CommunityScripts? I think it'd be easier packaging as well, but y'all can decide that, it's your repo after all 🙂 worst case I can just create a new repo")

So what does DD think? It's no hard feelings if the consensus is that it's better outside this repo (despite it being a Stash plugin), but I'd like DD's opinion even if it's simply "close without comment" haha :) Thanks!

DogmaDragon:
I think it would be a better fit for it's own index for maintanance purposes, but I have no issues with merging it. We already have plugins that depend on external servers or are companions to external tools. 

One thing I would request is to add to the plugin description that it requires StashPlexAgent.bundle, this way it will be visible to everyone that looks at the plugin through the package manager.

feederbox826:
Didn't realize https://github.com/Darklyter/StashPlexAgent.bundle/pull/39 was merged, Once the description fix is added I'll merge

HarryS:
Maintenance is not a big worry for me, as it's simple and stable, but I of course take up responsibility for it if something on Stash side would break it (and if something from Plex side breaks it, it means that Plex has most likely decided to kill their entire API and thus I'd request to remove the plugin haha)

Does it look good now? Feel free to modify wording if you have a better flow, cheers :)

PR #573, opened 2025-06-03
Add VideoBanner Plugin

philMorel:
I've made a new plugin called VideoBanner.
VideoBanner displays a random preview video sourced from one of the scenes associated with the current entity. This video plays as a background in the main header area of the entity's detail page (right where the item's data is shown).
On Tag detail pages, the video source changes depending on whether the Scenes, Markers, or Groups tab is active, showing relevant previews for each.
FYI, an AI helped draft the initial code for this. It's been tested on my local setup and works fine.


WeedLordVegeta420:
Code is indeed a bit messy but it does work and does what you wanted it to do.

PR #575, opened 2025-06-06
[Video Chapter Markers] Fix on update trigger

you-cant-see-me:
from what i can see, on stash 0.28.1 (latest stable release at time of writing) `Scene.Update.Post` gets called with only `args.hookContext` in json input. `args.mode` is only present for the `processAll` action when manually triggered in tasks section, and it is set like `args.mode == "processAll"`. the branch checking for `args.hookContext` was never reached as `args.mode` does not exist in these instances.


you-cant-see-me:
@DogmaDragon done :+1:

PR #581, opened 2025-06-13
[audio-transcodes] Fix paths with spaces and ffmpeg stderr flood

you-cant-see-me:
fixes audio transcodes with files that have spaces in them, forming a command using `%` (or any other string manipulation) is very dangerous at worst and wrong at best, this job should be handled by an appropriate library for running commands which can separate command arguments appropriately, aka subprocess.

also changes the ffmpeg command for audio transcodes to log fatal errors, otherwise the whole transcode log would show up as errors in stash even when it succeeded. also removes unnecessary logging of the whole stash configuration dict which can be several pages worth of scrolling while looking at logs.


Tweeticoats:
this looks good

PR #584, opened 2025-06-21
Added PythonDepManager to avoid people not installing dep

Nonoty:
Added PythonDepManager to avoid people not installing dep

### DIFF
--- a/plugins/DateParser/date_parser.py
+++ b/plugins/DateParser/date_parser.py
@@ -1,6 +1,7 @@
 import sys, json
 from PythonDepManager import ensure_import
 ensure_import("dateparser>=1.2.1")
+ensure_import("stashapi:stashapp-tools>=0.2.58")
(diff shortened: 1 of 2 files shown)


Nonoty:
Because its been years since i first made this plugin and i consider it to be stable enough now that it can be 1.0

dbuttric:
If I'm using the docker container, how do I do this? I applied the diffs, now I'm getting that I need to install that PythonDepManager.
When I try to execute pip in the container, I get the error about not having virtual environment.
Can you tell me where the venv is so I can source it, and install packages?
Thanks for the help.

NotMyMainUser:
PythonDepManager should be installed automatically when you install the Plugin via your stash UI. If that didn't work for some reason you can install it from the plugin section in stash

NotMyMainUser:
I'll try to fix these issues in PythonDepManager this weekend

PR #586, opened 2025-06-24
[nfoSceneParser] new config options

deLbe321:
Added possibility for nfoSceneParser plugin to configure:
- custom name of the nfo file
- name of the user rating field
- user rating multiplier
- source for tags

Additionally added:
- missing studio information on scene update
- "folder" as possible name for scene image

### DIFF
--- a/plugins/nfoSceneParser/config.py
+++ b/plugins/nfoSceneParser/config.py
@@ -10,6 +10,11 @@ dry_mode = False
 # ! Not yet implemented. Currently, only "with files" is supported
 nfo_location = "with files"
 
+# By default the plugin will look for an nfo based on file name:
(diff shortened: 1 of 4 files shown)


Gykes:
From an initial review it looks good. I will test and review more thoroughly when I get home from work today.

deLbe321:
> From an initial review it looks good. I will test and review more thoroughly when I get home from work today.

I will greatly appreciate that. Although I tried to not change the default behavior, the last thing I would like to do is to break something for existing users. Update process will overwrite existing `config.py` for users?

deLbe321:
Ok so just in case I added some protection to make sure the new version works with old config file that doesn't have these new properties.

Gykes (approved):
Looks good

PR #590, opened 2025-07-14
Visage update On behalf of cc1234

Nonoty:
Small update to visage so that it works with latest backend
On behalf of cc1234


DogmaDragon:
What do you by "on behalf of cc1234"?  Does he commit to maintain the plugin going forward again or are instead you willing to become maintainer on record?

Nonoty:
He continues to maintain the backend updated. He is even improving it. As for the frontend, i dont have the source code (svelte pre compile) but its similar as my LocalVisage. So if it breaks, both do. 
But ideally we would need someone to rewrite the front, that would be the best. But so far the front hasnt break in 2y so i think we should be fine

Nonoty:
ah damn. Made a mistake. I though this was merged already

PR #607, opened 2025-08-27
Fix create_movie() deprecation in TMDBMarkers Plugin

bunkmate5127:
**Functional fix:**
- Replaced  deprecated call to stash's old function "create_movie()" with its new function "create_group()". Will remove the deprecation warning from the stash log.

**Tidying up the code:**
- Changed all variable and function names from referring to 'movies', to referring to the new terminology of 'groups' for consistency with the stash call. ('m'→'g',  'movie'→'group', 'mov'→'grp')
- Note: no change to moniker 'm' when used to represent a short-form for "marker".

### DIFF
--- a/plugins/TPDBMarkers/tpdbMarkers.py
+++ b/plugins/TPDBMarkers/tpdbMarkers.py
@@ -179,7 +179,7 @@ stash = StashInterface(FRAGMENT_SERVER)
 config = stash.get_configuration()["plugins"]
 settings = {
     "disableSceneMarkerHook": False,
-    "createGroupFromScene":True,
+    "createMovieFromScene":True,
     "addTPDBMarkerTag": False,
     "addTPDBMarkerTitle": False,
     "runOnScenesWithMarkers": False,


Tweeticoats:
I would keep the setting as createMovieFromScene as that is what the yaml has as the setting and that is the setting to toggle in the ui. I would recommend you update the yaml file and change the description of the setting.

bunkmate5127:
> I would keep the setting as createMovieFromScene as that is what the yaml has as the setting and that is the setting to toggle in the ui. I would recommend you update the yaml file and change the description of the setting.

Does stash have no viable migration pathway for variable names in plugin yaml files?

bunkmate5127:
I've made these changes that you've requested (I overlooked the yaml, thanks for pointing it out). 

My preferred solution would be to have instead of this regression to "createMovieFromScene" name instead have a migration. But I don't know what facility or schema the wider stash plugin system has (if any) to migrate historical plugin settings to new names. I accept that this might be non-trivial.

bunkmate5127:
...When a man can't even `ctrl+h` properly XD.

Sorry for the trouble, should be good now.

Issue #611, opened 2025-09-02
[TPDBMarkers] KeyError: 'groups'


Splash4K:
Getting this error when tpdbmarkers runs. Currently on stash `v0.28.1`

```
time="2025-09-02 14:53:10" level=error msg="[Plugin / The Porn DB Markers] Traceback (most recent call last):"
time="2025-09-02 14:53:10" level=error msg="[Plugin / The Porn DB Markers]   File \"/root/.stash/plugins/community/TPDBMarkers/tpdbMarkers.py\", line 206, in <module>"
time="2025-09-02 14:53:10" level=error msg="[Plugin / The Porn DB Markers]     processScene(scene)"
time="2025-09-02 14:53:10" level=error msg="[Plugin / The Porn DB Markers]   File \"/root/.stash/plugins/community/TPDBMarkers/tpdbMarkers.py\", line 64, in processScene"
time="2025-09-02 14:53:10" level=error msg="[Plugin / The Porn DB Markers]     for g in data[\"groups\"]:"
time="2025-09-02 14:53:10" level=error msg="[Plugin / The Porn DB Markers]              ~~~~^^^^^^^^^^"
time="2025-09-02 14:53:10" level=error msg="[Plugin / The Porn DB Markers] KeyError: 'groups'"
time="2025-09-02 14:53:10" level=error msg="Plugin returned error: exit status 1"
```

I believe all of my packages should be up to date as well

```
Package            Version
------------------ ---------
requests           2.32.3
stashapp-tools     0.2.58
```

Splash4K:
Confirming this issue is still appearing after the latest tpdbmarkers update (0.4.1-377ba6d 9/3/2025 2:23:49 AM)

```
time="2025-09-02 22:26:31" level=info msg="Updating package timestampTrade"
time="2025-09-02 22:26:36" level=info msg="Finished updating packages"
time="2025-09-02 22:26:36" level=info msg="Updating package TPDBMarkers"
time="2025-09-02 22:26:36" level=info msg="Finished updating packages"
time="2025-09-02 22:28:36" level=error msg="[Plugin / The Porn DB Markers] Traceback (most recent call last):"
time="2025-09-02 22:28:36" level=error msg="[Plugin / The Porn DB Markers]   File \"/root/.stash/plugins/community/TPDBMarkers/tpdbMarkers.py\", line 208, in <module>"
[...]

PR #633, opened 2025-11-16
Fix groups in TPDBMarker 

donlothario:
This should fix the bug retrieving movies. #611


plugins/TPDBMarkers/tpdbMarkers.py
```diff
@@ -61,7 +61,7 @@ def processScene(scene):
                     # skip if there is already a group linked
                     if settings["createMovieFromScene"] and len(scene.get("groups", [])) == 0:
                         groups=[]
-                        for g in data["groups"]:
+                        for g in data["movies"]:
```
axpepper: data.get("groups") is a better solution. It currently fails when a user has no groups ( or something, I'm not entirely sure what evoked a failure case but that fixes it )

PR #643, opened 2025-12-11
SFWSwitch: blur tagger scene image popup

Clasp8775:
In Tagger mode when you get a result for a scene (for example by clicking Scrape by fragment), the image is not blurred. 

By adding the missing CSS for .scene-image it now works as expected.


discourse-stashapp:
This pull request has been mentioned on **Stash**. There might be relevant details there:

https://discourse.stashapp.cc/t/sfw-switch/4658/4

PR #698, opened 2026-04-10
[Extra Performer Info] Bugfixes, setting user agent

Tweeticoats:
Bugfixes:
The main problem is we need to set a user agent for the requests or they will all be blocked.
Added a few extra fields and better handling where the person will receive the same award multiple times


discourse-stashapp:

https://discourse.stashapp.cc/t/extra-performer-info/1332/5

PR #700, opened 2026-04-13
Add Group Details plugin for Group Cards

Stash-KennyG:
## Summary
Add the `GroupDetails` plugin for group cards with computed duration and average vertical resolution metrics from GraphQL scene data

## What It Adds

- **Date line:** appends total duration (`H:MM:SS`) to the right side of the date row.
- **Chip list:** appends a resolution chip (PNG badge) to the end of `.card-popovers`.
- **Tooltips:** duration and resolution both expose tooltips with more detailed information.

## Overview
image

## Data Source

Metrics are computed in-browser from GraphQL `findGroup` scene data (`id`, `title`, `files { duration height }`, `groups { group { id } scene_index }`).

## Scene Filtering

The goal of the plugin is to include the feature portion of the attached scenes.  Often individuals select -1 or 99 for things like trailers or BTS.  This filter can be disabled by ticking **include all scenes**.  When disabled (default), scenes are included only if `scene_index` is:

- `null`, or
- an integer in `0..89`

Exception: if the group has exactly **one scene**, scene-index filtering is bypassed for that group.

When **Include all scenes** is enabled, all returned scenes are included regardless of `scene_index`.

image

## Sorting

Duration tooltip scene lines are sorted by:

1. `scene_index` ascending (`null` sorts as `90`)
2. duration descending
3. scene `id` ascending (stable tie-break)

## Duration Metric

- Uses each included scene's **max file duration**.
- Card value is total duration displayed as `H:MM:SS`.
- Tooltip lists every included scene as:
  - `N. Title H:MM:SS` when `scene_index` is present
  - `Title H:MM:SS` when `scene_index` is null

## Resolution Metric

Average resolution uses vertical pixels (`height`) from each included scene's tallest file:

- For groups with **exactly one total file**, the duration gate is bypassed.
- Otherwise, only scenes with `duration > 360` are eligible. **Goal**: Exclude sub-6 minute scenes to filter trailers etc.
- Resolution average is `round(sum(height) / count)`.
- Tooltip format is:
  - `Resolution Average: <N>p`
  - or `Resolution Average: —` when no eligible average exists.

Resolution chip empty/dash behavior:

- If there are no files (or single-file case with unusable height): render nothing.
- If `totalFileCount > 1` and no eligible average: render `—`.

## Resolution Badge Mapping

The plugin picks a PNG badge using a 2% tolerance (`>= 98%` of target resolution):

- `< 234` -> `144p.png`
- then highest match from:
  - `240`, `360`, `480`, `720`, `1080`, `1440 (2k)`, `2160 (4k)`, `2880 (5k)`, `3160 (6k)`, `4320 (8k)`

## Test plan
- [x] open `/groups` and verify duration appears on the date line, right-aligned
- [x] verify resolution chip is appended as the last popover chip and tooltip reads `Resolution Average: <N>p` or `—`
- [x] confirm scene filtering/sorting behavior in duration tooltip (`scene_index` ordering with null handling and secondary duration sort)
- [x] run `bash plugins/GroupDetails/build.sh` and confirm `images.js` is regenerated from `assets/*.png`
- [x] hard refresh Stash UI and confirm embedded image badges render without plugin asset URL dependence

AI assisted. Human Verified


plugins/480p.png
DogmaDragon, line 1: This looks to be a rogue file?

plugins/GroupDetails/build.sh
DogmaDragon, line 1: This looks to be a production file?

Stash-KennyG: Good question.
I guess I need a documentation addition.
The icons on the plugin work by calling images.js for a mapped Base64 encoded file.  Those files you see in the build and assets folder allow the re-generation/swapping of base files.  So if you wanted your own logos, you can do that with the build.sh file.

Stash-KennyG:
@DogmaDragon I have also included a second feature into this PR that includes the performer overlay as well.
Like the scene card, I've added the performers in each movie card.
I think it's ready for a merge... let me know on Discord if I missed anything.

PR #18, opened 2021-10-16
renamerOnUpdate: segment config options into different file

jthrow0451:

PR #704, opened 2026-04-21
[mobileWallLayout] Add play-on-visibility and ordered loading

speckofthecosmos:
## Summary

Two mobile-first behaviors on top of the existing layout fix, and a
display-name rebrand to `scrollFeed` to reflect the broader feature set.

- **Play-on-visibility.** An `IntersectionObserver` plays each `<video>` at
  10% visibility and pauses it when it leaves the viewport. Concurrent
  playback caps at 2–3, which keeps iOS Safari comfortably under its
  ~20-simultaneous-`<video>` decoder ceiling. On a 20-card wall page
  (Stash's minimum per-page size), the unchanged plugin leaves every
  preview autoplay firing at once — this PR fixes that.
- **DOM-ordered load queue.** On page mount, Stash fires parallel fetches
  for every `<video src>`. Over cellular that splits the uplink 20 ways
  and no video is playable for many seconds. The queue cancels those
  fetches via `removeAttribute('src') + video.load()` and re-issues them
  top-down, 2 at a time. Advances on `canplay` or a 500ms fallback — the
  top clip is playable fast on cellular, and the whole page is in-flight
  within ~5 seconds so moving into a weaker-signal area doesn't leave the
  bottom of the page with zero bytes.
- **Rebrand.** Display name in the YAML is now `scrollFeed` with an
  updated description. The plugin ID stays `mobileWallLayout` so existing
  installs upgrade cleanly.

## Internals

- File wrapped in an IIFE to keep top-level identifiers out of global
  scope (Stash loads all UI plugins into the same document).
- URL matching uses `window.location.pathname` (exact match on `/images`
  and `/scenes/markers`) rather than `href.includes`.
- Stale-timer safety: `beginLoading` captures the current `_loading` Set
  in its closure so any `setTimeout` advance that fires after
  `deactivateVideoBehavior` nulls plugin state detects the mismatch and
  no-ops.
- All non-iteration references use `WeakMap`/`WeakSet` so video GC isn't
  blocked when React unmounts elements.

## Tuning knobs

Constants at the top of the load-queue section:

| Constant | Default | Effect |
|---|---:|---|
| `threshold` (IntersectionObserver) | `0.1` | Lower = more continuous scroll; higher = stricter per-clip focus. |
| `_MAX_CONCURRENT_LOADS` | `2` | Higher = full-page finishes sooner; lower = top clips get more bandwidth. |
| `_LOAD_ADVANCE_MS` | `500` | Short = tail starts fetching sooner; long = top clips get more solo time. |

## Test plan

- [x] iPhone 16 Pro Max against a home-NAS Stash instance over remote
  cellular: 20-card marker wall plays 2–3 concurrent, top video canplay
  in ~1s, whole page in-flight by ~5s, scroll to video #18 shows bytes
  already accumulating.
- [x] Airplane-mode mid-scroll: already-loaded clips continue playing;
  entering a not-yet-loaded clip shows a queued state and resumes once
  back online.
- [x] Navigate away from `/scenes/markers` and back: observers tear down
  and re-attach cleanly, no leaked style tag or decoder stampede.
- [x] `validator/validate.js` passes locally; YAML schema-valid.

Internal plugin ID is unchanged (`mobileWallLayout`), so users with the
current published plugin will transparently upgrade.

### DIFF
--- a/plugins/mobileWallLayout/README.md
+++ b/plugins/mobileWallLayout/README.md
@@ -1,29 +1,68 @@
-# Mobile Wall Layout
+# scrollFeed (mobileWallLayout)
 
-https://discourse.stashapp.cc/t/mobile-wall-layout/6160
+Turns Stash's **Markers** (`/scenes/markers`) and **Images** (`/images`) wall
+into a scrollable mobile feed — full-width single-column layout, video
+play-on-visibility, and DOM-ordered loading that keeps the feed watchable
+over cellular and degrading connections.
 
-Makes the wall-mode gallery render as a single full-width column on mobile
-devices, on the **Markers** (`/scenes/markers`) and **Images** (`/images`) pages.
+The plugin is published under the filename `mobileWallLayout` (and that's
+still the internal ID, so existing installs upgrade cleanly). The display
+name it presents in Stash's Plugins panel is `scrollFeed`.
 
-By default, Stash's wall mode uses `react-photo-gallery`, which calculates
-`position: absolute` offsets for a multi-column brick layout. On small screens
-this produces items that are too small to comfortably tap and browse. This
-plugin overrides those offsets so each item spans the full width of the screen,
-making marker previews and images easy to scroll through on a phone.
+## What it does
 
-## Behaviour
+1. **Full-width single-column layout on touch devices.** By default, Stash's
+   wall uses `react-photo-gallery`, which calculates `position: absolute`
+   offsets for a multi-column brick layout. On phones those offsets produce
+   items that are too small to tap through comfortably. The plugin injects
+   a `<style>` tag wrapped in a `@media (pointer: coarse)` query to override
+   them. Touchscreens get the mobile feed; desktop and mouse-driven
+   viewports are untouched.
 
-- Applies only on **touch-screen devices** (`pointer: coarse`) — correctly
-  targets phones and tablets without triggering on narrow desktop browser windows.
-- Activates and deactivates automatically as you navigate between pages.
-- Has no effect on desktop or mouse-driven viewports.
+2. **Play-on-visibility.** Stash marks marker previews with `autoPlay`, so
+   a 20-card page can fire 20 simultaneous playbacks — iOS Safari bogs down
+   past its ~20-`<video>` decoder ceiling. An `IntersectionObserver` plays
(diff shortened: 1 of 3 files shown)


plugins/mobileWallLayout/mobileWallLayout.yml
```diff
@@ -1,10 +1,10 @@
-name: Mobile Wall Layout
+name: scrollFeed
 description: >
-  On the Markers and Images pages, forces the wall-mode gallery to render as a
-  single full-width column on mobile devices (phones in portrait or landscape,
-  up to ~960px wide). Tablets and desktops are unaffected.
-version: 1.0
-url: https://discourse.stashapp.cc/t/mobile-wall-layout/6160
```
DogmaDragon, line 7: Do not remove the link to community forum. We require it for all plugins published on this repository.

plugins/mobileWallLayout/README.md
```diff
@@ -1,29 +1,68 @@
-# Mobile Wall Layout
+# scrollFeed (mobileWallLayout)
 
-https://discourse.stashapp.cc/t/mobile-wall-layout/6160
```
DogmaDragon, line 3: Do not remove the link to community forum. We require it for all plugins published on this repository.

speckofthecosmos:
Apologies — the scrollFeed rebrand silently dropped the discourse URL from two places: the `url:` field in `mobileWallLayout.yml` was repointed to a GitHub tree URL, and the discourse link was removed from the top of the README. That breaks the convention every other plugin in this repo follows (the `url:` field is what the Stash Plugins panel surfaces as the support link), and it was my oversight.

To be upfront: this PR was drafted with AI assistance (Claude Opus 4.7 — noted in the first commit's `Co-Authored-By` trailer). The rebrand/rewrite step is where the discourse URL was silently dropped, and I did not catch it on my review.

Just pushed b6ad2e3 restoring both:
- `mobileWallLayout.yml` → `url: https://discourse.stashapp.cc/t/mobile-wall-layout/6160`
- README → `Discussion:` link at the top

Adding a pre-PR check to my own routine so the `url:` field gets verified against the discourse convention before any future CommunityScripts plugin PR I open.

discourse-stashapp:

https://discourse.stashapp.cc/t/mobile-wall-layout/6160/3

PR #705, opened 2026-04-26
Embed Metadata Plugin for Stash

findmenow404:
This is a plugin for stash that will re-curse through your library and re-write the metadata in the MP4 and MKV files with the information from Stash.  Additional options could be added to prevent re-write of files that have already been performed, or a selection.

### DIFF
--- a/plugins/EmbedMetadata/README.md
+++ b/plugins/EmbedMetadata/README.md
@@ -1,0 +1,86 @@
+# Stash Metadata Embedder
+
+A [Stash](https://stashapp.cc) plugin that embeds scene metadata and cover art directly into MKV and MP4 video files using ffmpeg. No re-encoding — all operations are stream copy, so processing is fast and lossless.
+
+## Features
+
+- Embeds title, date, performers, studio, tags, rating, details, and URL into video file metadata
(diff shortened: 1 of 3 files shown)


DogmaDragon:
Stash keeps track of files by path and oshash. You do replace in-place, but embedding metadata changes file oshash.

For me to merge this, I would want for it to handle updating database with the new file oshash. Otherwise if someone decided to move these newly created files, they would be left untracked.

findmenow404:
Do you know of another way to embed the data without it changing the oshash, since it's based on the first and last 64K of the file? I was hoping to keep it matching to stashdb databases for the oshash, but more as a secondary backup embeded directly in the file.  I can add a auto re-scan and update documentation regarding the oshash change if needed.
Or possibly I'll update it to just export .nfo and .jpg files along side.

DogmaDragon:
> Do you know of another way to embed the data without it changing the oshash, since it's based on the first and last 64K of the file?

No, as it also takes the file size. See https://github.com/opensubtitles/oshash#algorithm

```
hash = file_size + sum_uint64_le(first_64KB) + sum_uint64_le(last_64KB)
```

PR #706, opened 2026-04-26
Add Info JSON Importer plugin

MaxProvider:
## Info JSON Importer

Adds a new plugin that imports metadata from yt-dlp .info.json files directly into Stash scenes.

### What it does

Many users have large video libraries downloaded with yt-dlp that have .info.json files saved alongside the videos. Until now there was no way to get that metadata into Stash without scraping the source sites again. This plugin reads the local .info.json files and imports directly from them.

For each scene in the Stash library, the plugin looks for a matching .info.json file next to the video file and imports:
- Title
- Date (upload_date)
- Studio (from uploader/channel, created if not exists)
- Tags (created if not exists)
- Description (into the Details field)

### Tasks

Two tasks are provided:
- **Dry Run** — previews what would be imported without making any changes
- **Import Info JSON** — performs the actual import

### Notes

- Written in Python using the Stash GraphQL API
- Includes a setup.sh script that asks for the API key and copies files automatically
- Tested on Stash v0.31.1 on Linux
- Works with Pornhub and XHamster .info.json files (and any other yt-dlp generated .info.json)

Developed by David Smith © 2026
david@maxprovider.net

### DIFF
--- a/plugins/infoJsonImporter/README.md
+++ b/plugins/infoJsonImporter/README.md
@@ -1,0 +1,156 @@
+# Info JSON Importer
+
+**Imports metadata from yt-dlp `.info.json` files directly into Stash.**
+
+Developed by [David Smith](mailto:david@maxprovider.net) · © David Smith 2026
+
+---
+
+## What It Does
+
+When you download videos with [yt-dlp](https://github.com/yt-dlp/yt-dlp), it can optionally save a companion `.info.json` file alongside each video containing metadata from the source site — title, upload date, tags, uploader name, description and more.
+
+This plugin scans every scene in your Stash library, looks for a matching `.info.json` file next to the video, and imports the following fields into Stash:
+
+| Field | Imported As |
+|---|---|
+| `title` | Scene title |
+| `upload_date` | Scene date |
+| `uploader` / `channel` | Studio (created if not exists) |
+| `tags` | Scene tags (created if not exists) |
+| `description` | Scene details |
+
+Studios and tags that do not already exist in Stash are created automatically during the import.
+
+---
+
+## Requirements
+
+- Stash v0.25.0 or later
+- Python 3 (accessible as `python3`)
+- pip
+- yt-dlp `.info.json` files saved alongside your video files
+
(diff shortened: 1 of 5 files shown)


DogmaDragon:
This looks to be better suited to be a scraper. Like this one for yt-dlp https://discourse.stashapp.cc/t/yt-dlp/5109.

MaxProvider:
> This looks to be better suited to be a scraper. Like this one for yt-dlp https://discourse.stashapp.cc/t/yt-dlp/5109.

Thank you for your comment.

This is not a scraper.

It is a plugin that reads already scraped info.json meta files to appened the contained data to matched videos within the stash database.

DogmaDragon:
As does the one I linked. It uses Stash scraper system instead of doing the same as a plugin.

DogmaDragon:
I would suggest hosting your own plugin source index for this one. We have a template you can use to quickly set it up. https://github.com/stashapp/plugins-repo-template.

MaxProvider:
There are two key distinctions here:

A scraper, which collects metadata.
A plugin, which adds that metadata to videos within the Stash database.

This plugin does not scrape data from the web. It simply adds pre-obtained info.json metadata to the StashDB, associating it with the corresponding video(s).

Currently there is no plugin available that carries out this task for those who already have info.json files or further downloaded them seperately.

For example, the current Pr0n-hub plugin does not work, it crashes, thus users are unable to automatically add the required meta data, this solves that by allowing users who have already scraped, or will further scrape the info.json files themselves, this plugin can then be used to attribute the contained data to the corresponding video(s).

Maista6969:
In the context to Stash, scrapers are just what you use to get metadata about your content into Stash: the source of the metadata does not need to be a website :slightly_smiling_face: 

There are scrapers that read the file metadata and also sidecar files similar to what you've implemented

MaxProvider:
> In the context to Stash, scrapers are just what you use to get metadata about your content into Stash: the source of the metadata does not need to be a website 🙂
> 
> There are scrapers that read the file metadata and also sidecar files similar to what you've implemented

There needs to be a clear definition of what a plugin does and what a scraper does within this context then, as it seems pretty ambigious between the two, from what I now understand, causing confusion for those who are unaware of the ambiguity.

Poor denfinitions.

PR #710, opened 2026-05-06
[SFW Switch] Fix: Corrected incorrect state when refreshing page

Servbot91:
## Fix

- Should correct the race condition causing the effect to apply when not selected, added better check for localstorage
- Added additional CSS setting for Ascension


discourse-stashapp:

https://discourse.stashapp.cc/t/sfw-switch/4658/31

Servbot91:
Need additional testing

Servbot91:
Should be good now

PR #715, opened 2026-05-23
feat(SFWSwitch): add Never Unblur on Hover setting

evolite:
Adds a new boolean plugin setting **Never Unblur on Hover** (`never_unblur`) to the SFW Switch plugin.

When enabled, all blurred content stays blurred even when hovered.

## Changes
- `sfwswitch.yml`: new `never_unblur` boolean setting
- `sfw.js`: read the new setting and inject/remove a `<style>` element that overrides the hover unblur rules when active


discourse-stashapp:

https://discourse.stashapp.cc/t/sfw-switch/4658/32

PR #718, opened 2026-05-29
Feat: Significant performance improvements across all mass taggers

Servbot91:
## Dev Notes

These changes provide on average about a 95% performance improvement to all 4 mass taggers. I have tested these individually and running concurrently without issue. Mass tagging galleries, images, and scenes has dramatically increased. The previous logic did not properly use the schema which caused the DB to get absolutely blasted with "slow query" messages in debug. 

If running these plugins concurrently, they would ping pong off each other causing loops further running away with your DB's performance. You also had the issue of the scripts reiterating over files they should be ignoring which also contributed to unnecessary load. If your database size is significant or you have large galleries this could bring your DB to its knees and would basically require you to run these tasks over night.

For using the 'Tag all galleries from images' feature, I have gone from over 2 days ETA to 40 minutes across 2.5 million images.
'Tag Images with performer' tags of 5k+ performers has been reduced to less than 30 minutes down from 7 hours.
'Tag all images from galleries' (80k + galleries) is down from 14 hours to 40 minutes.
'Tag all scenes' (20k+) went from 4 hours to 20 minutes.

These changes maintain core logic while also improving performance. User performance will vary, but at the very least the scripts are running significantly more optimal than they were before.

-Sakoto

---

These images demonstrate that no more slow queries display under load with all taggers running in tandem and a mass tagger running. 

Screenshot 2026-05-28 222025

Screenshot 2026-05-28 221850

Screenshot 2026-05-28 205124

## Changes
- Galleries to images now tags studios down to images
- All mass tagging tags are significantly faster (95% performance gain)
- Taggers no longer loop and piggy back off each other when used at the same time causing a performance runaway
- Database no longer crawls when performing a mass tag allowing users to continue to use their DB (and even tag individually) without destroying DB performance
- Taggers better utilizing stash schema
- Fixed infinite loop bug with pagination
- Taggers ignore entities that do not need updating
- More informative log messaging
- Better Debug info
- Better hook management
- Loop safety checks

### DIFF
--- a/plugins/tagGalleriesFromImages/tagGalleriesFromImages.py
+++ b/plugins/tagGalleriesFromImages/tagGalleriesFromImages.py
@@ -2,13 +2,22 @@ import stashapi.log as log
 from stashapi.stashapp import StashInterface
 import sys
 import json
+import time
+
+GALLERY_PAGE_SIZE = 150
+tag_cache = {}
+
 
 def processAll():
     exclusion_marker_tag_id = None
-    if settings["excludeWithTag"] != "":
-        exclussion_marker_tag = stash.find_tag(settings["excludeWithTag"])
-        if exclussion_marker_tag is not None:
-            exclusion_marker_tag_id = exclussion_marker_tag['id']
+    if settings["excludeWithTag"]:
+        if settings["excludeWithTag"] in tag_cache:
+            exclusion_marker_tag_id = tag_cache[settings["excludeWithTag"]]['id']
+        else:
+            exclusion_marker_tag = stash.find_tag(settings["excludeWithTag"])
+            if exclusion_marker_tag:
+                exclusion_marker_tag_id = exclusion_marker_tag['id']
+                tag_cache[settings["excludeWithTag"]] = exclusion_marker_tag
     
     query = {
         "image_count": {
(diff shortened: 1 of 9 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/tag-images-from-galleries/6024/5

DogmaDragon:
Could you split this into separate PRs? It's updating 3 plugins from 3 different authors and adding a new one on top of that.

Servbot91:
I mean, i could. Though im not sure what difference it makes outside of polluting commit history for plugins that have been practically abandoned and creating more work for me simply for the sake of it.

Sent from Mobile🚄

-------- Original Message --------

DogmaDragon:
1. It makes review process harder. 
2. If there are needed changes, it will block the whole PR even if only one plugin is affected. 
3. We give original authors a chance to review PRs targetting their plugins, which will delay the merge of whole PR instead of only affected plugin, 
4. I prefer clear and conside PRs for adding new plugins as it makes tracking commit history a lot easier down the line and to avoid license issue with <https://github.com/Servbot91/StashAPP-TagImagesFromGalleries/tree/main/plugins/tagImagesFromGalleries> version which doesn't have any license.

Servbot91:
If that is how you believe things should be then that is fine. I will go ahead and maintain these on my own.

The bureaucracy involved with submitting changes to this repo seems to continuously create more work than its worth for me. For plugins with inconsistent or no quality control that are basically abandoned, it seems much easier for me to address these problems without a middle man. You can reject the pr.

Sent from Mobile🚄

-------- Original Message --------

DogmaDragon:
> The bureaucracy involved with submitting changes to this repo seems to continuously create more work than its worth for me.

We don't see having license complience and giving original authors an opportunity to review changes before they are pushed as bureaucracy, but it's fair if that's too much for you.

PR #721, opened 2026-05-30
Significant performance improvement across tagGalleriesFromImages

DogmaDragon:
Related PR https://github.com/stashapp/CommunityScripts/pull/718


discourse-stashapp:

https://discourse.stashapp.cc/t/tag-galleries-from-images/3904/3

plugins/tagGalleriesFromImages/tagGalleriesFromImages.py
```diff
@@ -2,13 +2,22 @@
...
+
 
 def processAll():
     exclusion_marker_tag_id = None
-    if settings["excludeWithTag"] != "":
-        exclussion_marker_tag = stash.find_tag(settings["excludeWithTag"])
-        if exclussion_marker_tag is not None:
-            exclusion_marker_tag_id = exclussion_marker_tag['id']
+    if settings["excludeWithTag"]:
+        if settings["excludeWithTag"] in tag_cache:
```
spaceyuck, line 14: I really don't get the point of the `tag_cache`. `processAll()` gets invoked once, so there are no repeat lookups of the same thing to cache. And even then, it's a single tag lookup by name at worst. Should be removed if pointless, just added complexity.

plugins/tagGalleriesFromImages/tagGalleriesFromImages.py
```diff
@@ -18,95 +27,137 @@ def processAll():
...
-        log.progress((i / total_count))
+    try:
+        total_count = stash.find_galleries(f=query, filter={"page": 1, "per_page": 1}, get_count=True)[0]
+    except Exception:
+        total_count = 0
+
+    processed = 0
+    page = 1
+
+    while True:
```
spaceyuck, line 44: Not a huge fan of while loops without an explicit abort condition if there's an option to have one. Any reason `processed < total_count` is not feasible anymore?

plugins/tagGalleriesFromImages/tagGalleriesFromImages.py
```diff
@@ -18,95 +27,137 @@ def processAll():
...
+
+    existing_tag_ids = {t['id'] for t in gallery.get('tags', [])}
+    existing_perf_ids = {p['id'] for p in gallery.get('performers', [])}
+    existing_studio_id = gallery.get('studio', {}).get('id') if gallery.get('studio') else None
+
+    missing_tags = all_found_tag_ids - existing_tag_ids
+    missing_perfs = all_found_performer_ids - existing_perf_ids
+    
+    target_studio_id = list(all_found_studio_ids)[0] if all_found_studio_ids else None
+    studio_missing = bool(target_studio_id and (target_studio_id != existing_studio_id))
```
spaceyuck, line 108: Performers and tags are additive changes (so manually added tags and performers remain unchanged), but the studio gets overwritten if already set (manually or otherwise). This seems worth talking about, and maybe should be a setting to be enabled if wanted behavior by user.

plugins/tagGalleriesFromImages/tagGalleriesFromImages.py
```diff
@@ -18,95 +27,137 @@ def processAll():
...
+
+    if not missing_tags and not missing_perfs and not studio_missing:
+        return
+
+    update_data = {"ids": [gallery_id]}
+    if missing_tags and all_found_tag_ids:
+        update_data["tag_ids"] = {"mode": "ADD", "ids": list(all_found_tag_ids)}
```
spaceyuck, line 115: Why are all tags and performers added again, if the missing IDs are already in `missing_tags` and `missing_tags`? Seems like pointless extra work for the server.

plugins/tagGalleriesFromImages/tagGalleriesFromImages.py
```diff
@@ -18,95 +27,137 @@ def processAll():
...
+    hook_type = hook.get("type", "")
+
+    # --- GALLERY HOOK HANDLER ---
+    if hook_type in ["Gallery.Update.Post", "Gallery.Create.Post"]:
+        gallery = stash.find_gallery(hook['id'], fragment="id title code organized tags { id name } performers { id } studio { id }")
+        if gallery:
+            processGallery(gallery)
+
+    # --- IMAGE HOOK HANDLER ---
+    elif hook_type in ["Image.Update.Post", "Image.Create.Post"]:
```
spaceyuck, line 155: These image hooks are not declared in the manifest, and will never be called.

Also, should these be active by default? Every hook invocation already makes bulk update operations slower, and the images are not the ones getting changed, so I'd probably add a setting that has to be enabled to be more than a NO-OP

spaceyuck (changes requested):
This seems like a lot of code change noise for a few smallish actual changes:
* paged gallery loading (fair enough, didn't expect people to have that many galleries)
* studio gets pushed to gallery too
* hook into image changes (incomplete)

PR #723, opened 2026-06-07
fix(miscTags): Log config as debug to prevent spam

Larsluph:


discourse-stashapp:

https://discourse.stashapp.cc/t/misc-tags/1383/4

Tweeticoats:
sounds good

PR #725, opened 2026-06-07
[tagImagesWithPerfTags] optiinally exclude tags with "Ignore Auto-Tag" set

spaceyuck:
Makes sense to count this inheritance of tags as auto-tagging, at least optionally. Also very helpful, if you have lots of automatically generated performer tags (like from the performer body calculator plugin, or the Advanced Rating System Plugin), to keep these automatic performer tags from spilling all over (at least it puts you in control who will be ignored for inheritance).

(this is the same change as #724, just for images)


discourse-stashapp:

https://discourse.stashapp.cc/t/tag-images-from-performer-tags/2059/5

PR #730, opened 2026-06-18
Upgrading Stashai to ONNX endpoint

Drewlius:
#### AI USAGE
Opencode was used to help me understand the syntax of the API requests to the Stashai ONNX Model endpoint, the GRADIO 5 documentation was fed into it, HF API Request documentation for JS was fed in, the old stashai.js was fed in, and the contents of my IDE was fed in, this was done to query when I had questions regarding the documentation and helping to troubleshoot non 200 OK Responses from the HF API endpoint. JavaScript is also fairly new to me so it was useful and time saving for me to get help with particular JS Syntax and plain English translations of functions and definitions contained within the original .JS file. The code was all tested and verified to be functioning properly and returning both Tags and Markers and successfully attaching them to the video file and adding those Tags/Markers to the DB properly. There may be issues still present. I am not an expert at web development or Javasript and have spent most of my time in programming writing Python and bash scripts for Linux, Like i stated though the script is working and returning responses from the correct ONNX API endpoint. If you review and see something is wrong don't hesitate to point it out. I will spend the time to fix it to the best of my abilities.


Drewlius:
Apologies about the original pull requests to the wrong branch of the project.

plugins/stashAI/stashai.yml
DogmaDragon: Need to bump the version.

Drewlius: Okay I fixed that issue

Drewlius (commented):
Verified to be a changed version per the maintainer directive. The version now reflects 1.0.3 compared to previously showing 1.0.2

DogmaDragon:
Can't get it to work. The request is being blocked.  
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://cc1234-stashtag-onnx.hf.space/gradio_api/call/predict_tags. (Reason: CORS request did not succeed). Status code: (null).
```

Drewlius:
Okay give me a few. It was working 2 weeks ago when I submitted the change.

> Can't get it to work. The request is being blocked.
> 
> ```
> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://cc1234-stashtag-onnx.hf.space/gradio_api/call/predict_tags. (Reason: CORS request did not succeed). Status code: (null).
> ```

DogmaDragon:
Got it to work. My bad for not hard refreshing. 

Upon testing it looks to be noticeably slower. Ignoring first request, the subsequent requests are 2-5 times slower than non-ONNX endpoint. Results looks to be same, so not sure what is the benefit of this bigger and better model.

Drewlius:
I get more returns with better results based on the testing I did, with a larger tag database to pull from. From what I could find for the reason it wasn't working with the other code it appears some back end Gradio changes. The ONNX Space's 'requirements.txt' pins 'gradio>=5.0.0' no minor-version lock. On container rebuild (cold start, cache eviction, or infrastructure update), pip resolves to whatever latest Gradio 5.x satisfies the constraint. The '/gradio_api/call/{event_id}' SSE path was never the correct endpoint for a queue-enabled Gradio 5.x space. it just so happened to work with whatever version or cache state the container had when I tested it previously. I am about to update the files now. Tested several times and verified functional. if it still fails to load it could be due to cold start of the huggingface server. But i definitely seem to get more results with a larger tags pool to pull from with this version. Maybe try again with the newly pushed commit.

> Got it to work. My bad for not hard refreshing.
> 
> Upon testing it looks to be noticeably slower. Ignoring first request, the subsequent requests are 2-5 times slower than non-ONNX endpoint. Results looks to be same, so not sure what is the benefit of this bigger and better model.

Drewlius (commented):
Tested on my machine on several clips and verified to be functional. I would recommend going to the [ONNX](https://huggingface.co/spaces/cc1234/stashtag_onnx/) url jsut to verify that the model is up and running as HuggingFace will spin the server down after a certain amount of time with no activity. This can cause the stashai functionality to be lost until the server is loaded. I was able to get new tags to appear that I was not getting returned to me with the old non-ONNX endpoint as well as the accuracy of the tags seemed more precise (anecdotally) than they did prior to the change. Your mileage may vary. Both the .js & .yaml need to be added to your code base for change to take effect, as well as a hard browser cache reload, and refreshing the plugins via the app interface. I would also recommend giving the application a hard restart as well as the browser

I would order it in this way: Plugin refresh in application interface plugins setting --> Hard application full restart --> clear browser cache and restart the browser then reconnect to your hosted endpoint. Hopefully this will be the final needed change to this *fingers x crossed*. Keep me posted with any further issues you may run into.

--P.S. the 'agents.md' file is not strictly necessary and was linked to on the models landing page for use if an agent wants to be integrated to work with the model. I included it in the off event someone would want integration with an agent of theirs. This can be included or not at the final verification and publishing stage, if we are able to get to that point.

DogmaDragon:
I feel like the better approach would be release this version as a new plugin instead of replacing existing one as the speed and reliablity is not universally better. Most end users will have to idea that they need to go the endpoint to make HF to restart the space for their pugin to work.

Also, @cc1234475 recently released a new space [stashtag_onnx](https://huggingface.co/spaces/cc1234/stashtag_onnx) which uses ONNX model by default, not sure how that changes the implementation.

Drewlius:
Okay, well on his main profile landing page it appears that the old StashTag HF space was removed and the stashtag_onnx space is the only one available now, check [here](https://huggingface.co/cc1234). I am also assuming that the space must not have the cold start issue any longer as navigating to the page today appeared with the space completely loaded and ready to go.

discourse-stashapp:

https://discourse.stashapp.cc/t/stash-ai/1392/19

PR #733, opened 2026-06-27
fix(RenameFile): prevent scene title freezing on Next navigation

### LINKED ISSUE #551: Rename file plugin causes video title to not update when skipping to next video
**Describe the bug**
When skipping to the next video, the title displayed on the page still shows the first video. The browser tab title shows the correct title and all other metadata about the video correctly updates. This happens on both the local Windows server browser as well as mobile browsers. Restarting Stash does not fix the problem. This problem didn't exist. 

I didn't realize this was a Rename File plugin issue until entering this bug report and put more thought into it.

**To Reproduce**
0. Install rename file plugin
1. Go to play any video or play random.
2. Click on play, then click the next button. 
3. The title shows the prior video.
4. Click next button.
5. The title still shows the first video. 
 
**Expected behavior**
Update the title

**Screenshots**
If applicable, add screenshots to help explain your problem please ensure that your screenshots are SFW or at least appropriately censored.

**Stash Version: (from Settings -> About):**
v0.28.1

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Windows Chrome, Android Kiwi
 - Version Chrome 135.0.7049.115

**Smartphone (please complete the following information):**
 - Device: Pixel 7a
 - OS: Latest Android
 - Browser Kiwi

**Additional context**

garfolino:
## Problem

Fixes #551

When the RenameFile plugin is active on a scene detail page, pressing **Next** to navigate to another scene leaves the **previous** scene's title displayed. Every other field updates correctly; only the title is frozen. Disabling the plugin restores normal behavior.

**Root cause:** The scene title is rendered by a React-controlled `TruncatedText` component. `wrapElement()` replaced its contents with an `<a>` element via `element.innerHTML = ''`, which detaches the text node React tracks. On navigation, React writes the new title into that now-orphaned node, so the visible title never updates. This is a DOM-ownership conflict, not a caching issue (reproduces across browsers).

## Solution

Stop restructuring React's DOM. Instead:

- **Event delegation** — a single `document`-level `click` listener resolves the title via `event.target.closest('.scene-header div.TruncatedText')` and reads the text at click time. React renders/updates the title freely; we never touch its structure.
- **Non-destructive affordance** — the `.renamefile` class and tooltip are applied via `classList.add` / `setAttribute` only (no child mutation), so React's text node stays intact.
- **CSS** — `.renamefile` now uses `cursor: pointer` + hover underline on the title element itself.

All four interaction modes (click / Ctrl / Shift / Alt) are preserved through the unchanged `AppendTitleField`.

Version bumped `1.0.1` → `1.0.2`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


discourse-stashapp:

https://discourse.stashapp.cc/t/renamefile/1334/5

DogmaDragon (changes requested):
Need to bump the version in the .yml file.

PR #738, opened 2026-07-03
Add Metrics — comprehensive analytics dashboard

stash-metrics:
### What is it

A read-only analytics plugin that adds a Metrics tab to Stash's main nav
with 16 dashboards covering:

- Library overview (KPIs, duration/resolution/codec/rating distributions)
- Performer demographics (age pyramid, height, country, cup, body mods)
- Tag analysis (cloud, treemap, co-occurrence, cross-tab heatmaps, hierarchy)
- Performer collaboration network (force-directed graph with community detection)
- Time series (scenes/hours per month, tag popularity, debut timelines)
- Play patterns (day-of-week × hour heatmap, streaks, most/least watched, roulette, blind pick)
- Performer archetypes (k-means clustering + cosine-similarity lookup)
- Diversity / gap analysis (Shannon, Gini-Simpson, coverage gauges)
- Correlations (Cramér's V, Pearson, rating heatmaps, body-type calibration)
- Preference-driven Matches finder with weight sliders
- Fantasy Builder (reverse-engineers a preference profile from a seed set)
- Quality insights (health score, hidden gems, over/underrated z-scores,
  guilty pleasures, buried treasure, tag potency, rating drift)
- Yearly Wrapped slideshow, Bingo achievement grid, rotating nudge banner
- Tag Optimizer (near-duplicate + always-together detection, hierarchy audit)
- Cleanup tab (performer + scene metadata gap finder with chip filter, edit deep-links)

Also ships 5 runnable tasks: full report (JSON/CSV/HTML), cache refresh,
match finder, Year-in-Review, Excel export (multi-sheet .xlsx via SheetJS).

### Read-only

The plugin never writes to Stash's DB. Every mutation is done by the user
via the deep-linked native editor.

### LLM disclosure

Portions of this plugin were scaffolded with Claude (Anthropic) as a
pair-programming assistant. All code was reviewed and tested against a
real Stash instance before submission. I take full responsibility for
the code and its behaviour.

### Testing

- Smoke tests in `backend/test_aggregate.js` (offline, no Stash required)
- Manually tested end-to-end against a Stash instance with a
  multi-thousand-scene library
- All 16 dashboard tabs render on a fresh Stash install after running
  `Update Metrics Cache`

### Source

Full source (with commit history) also lives at
https://github.com/Stash-Metrics/stash-metrics-plugin


discourse-stashapp:

https://discourse.stashapp.cc/t/etrics-comprehensive-analytics-dashboard-cool-dashboard-to-visualize-who-you-like-has-a-fantasy-builder-to-define-your-preferences-etc/10789/1

DogmaDragon:
You should be able to map assets via plugin id. See
- https://docs.stashapp.cc/in-app-manual/plugins/#__codelineno-1-22:24
- https://docs.stashapp.cc/in-app-manual/plugins/#__codelineno-2-1:3

plugins/Metrics/Metrics.yml
```diff
@@ -0,0 +1,307 @@
...
+    defaultArgs:
+      mode: excel-export
+
+# Optional hooks — wired up if the corresponding setting is on. Stash will
+# still call us on every event; the script no-ops when the toggle is off.
+hooks:
+  - name: Refresh metrics after a scan
+    description: Recomputes the metrics cache once a scan task finishes.
+    triggeredBy:
+      - Scan.Complete.Post
```
DogmaDragon: This in nonsense.

plugins/Metrics/Metrics.yml
```diff
@@ -0,0 +1,307 @@
...
+    defaultArgs:
+      mode: hook
+      hook: scan
+
+  - name: Refresh metrics after metadata generation
+    description: Recomputes the metrics cache once metadata generation finishes.
+    triggeredBy:
+      - Generate.Complete.Post
```
DogmaDragon: Again, not a valid trigger.

plugins/Metrics/package-lock.json
DogmaDragon, line 1: Should be excluded.

plugins/Metrics/package.json
DogmaDragon, line 1: Should be excluded.

plugins/Metrics/.gitignore
DogmaDragon, line 1: Should be excluded.

plugins/Metrics/Metrics.yml
DogmaDragon, line 1: Multiline descriptions are way too verbose.

DogmaDragon, line 1: Please also remove line breaks.

DogmaDragon (changes requested):

stash-metrics:
Thanks for the review @DogmaDragon ,  pushed a fix commit addressing all four points:
- **Hooks removed.** `Scan.Complete.Post` / `Generate.Complete.Post` aren't valid Stash triggers; the block is gone entirely and the `refreshOn*` settings + backend hook branch went with it.

- **Descriptions collapsed** to single lines throughout `Metrics.yml`.

- **`ui.assets:` mapping added** so the backend cache is served directly as JSON under `/plugin/Metrics/assets/metrics-cache.json`. The old JS-bootstrap shim (`assets/metrics-cache.js`) is deleted; the dashboard now `fetch()`es the JSON.
- **`.gitignore`, `package.json`, `package-lock.json` removed** from the submission — those are dev-only files.
-
README (Hooks section dropped) and `docs/ARCHITECTURE.md` (cache lifecycle + v3.0 changelog entry) updated to match.

plugins/Metrics/Metrics.yml
stash-metrics: check main comments

DogmaDragon:
Was this ever tested by a human? 

Just from from a quick view there are some very obvious issues:
- 6.1 rating is getting treated as 4 stars.
- Setting options uses decimal numbers which UI can't input. 
- When default setting options are used, they are not noted. 
- Some charts can't be hovered on (most likely due to `min-height: 0`). 
- Long names are clipping outside the box boundries.

stash-metrics:
Small follow-up: fixed the ui.assets mapping key. Stash matches the keys against the path after stripping its own /plugin/<id>/assets/ prefix, so the correct mapping is /: assets — the previous /assets: assets double-prefixed the URL. Verified working on 0.31.1 (dashboard now loads the cache JSON directly).

stash-metrics:
> Was this ever tested by a human?
> 
> Just from from a quick view there are some very obvious issues:
> 
> * 6.1 rating is getting treated as 4 stars.
> * Setting options uses decimal numbers which UI can't input.
> * When default setting options are used, they are not noted.
> * Some charts can't be hovered on (most likely due to `min-height: 0`).
> * Long names are clipping outside the box boundries.

Hey DogmaDragon

thanks for checking, honestly I used a lot of AI but ran and tested it on my computer, it works quite nicely and tried to correct all the items, I run it daily against my own library (600 scenes, this is why i needed this, i like my search function), and my ratings are all full-star so the ceil() rounding never showed, and I never hit the decimal settings or had names long enough to overflow. Good catches, all fixed now.

now star buckets round to nearest instead of ceil (61/100 shows 3★ now), the two decimal settings are STRING since the number input only takes integers, defaults are written into every setting description, the hover issue was the min-height: 0 you spotted (canvas collapsed to no hit area, gave it a real min-height), and long names now ellipsize/wrap instead of overflowing.

DogmaDragon:
I don't believe that this was ever really human tested or reviewed. I didn't mention in the last message that **Settings** > **Logs** returned error for missing file to see if you would notice and fix it or just feed what I noted into an LLM and ignore the rest. The error is still present. 

```
error serving file <redacted>\Metrics\src\charts\files.js: open <redacted>\Metrics\src\charts\files.js: The system cannot find the file specified.
```

With that in mind, I don't feel comfortable merging this as it would probably be abandoned almost immediately and with it being 18k LOC there is no chance of someone else taking it over.

I would suggest creating your own index to host it instead. We have a pre-made template for that. See https://github.com/stashapp/plugins-repo-template.

stash-metrics:
Thanks for your time on checking and I understood your concern and makes me sure you really care for stash

Will keep playing with my plugin cause it was fun to work on it

Take care

PR #739, opened 2026-07-05
fix(tagScenesWithPerfTags): Speedup execution

Larsluph:
Check existing tags before apply on single scenes to prevent recursion limit hit on every update (2 calls max instead of 10) => 80-90% speedup


discourse-stashapp:

https://discourse.stashapp.cc/t/tag-scenes-from-performer-tags/1413/11

RubyDevil:
@Larsluph , this seemingly broke the following option:
image

PR #747, opened 2026-07-26
Visage v2.4.0

cc1234475:
### Visible loading state for face search
The face search trigger lives in the scene Operations dropdown, which closes the moment you click it. The scanning indicator lived on that button, so it was never visible, and the 2-3 second recognition call looked like nothing was happening.

The match modal now opens immediately with a scanning animation, and closes again if the request fails. The sprite modal and performer cards show their own loading state while a lookup is in flight.

### Details
- New `START_MATCH_SEARCH` action clears stale matches and opens the modal up front
- `FaceMatchModal` renders a scanning state instead of an empty result list
- `PerformerCard` and `SpriteResultModal` show progress during lookups
- CSS for the new scanning animations

### Tests
246 tests pass, including new coverage for the loading states in `FaceMatchModal` and `PerformerCard`. Type check is clean.

### CSP
`connect-src` unchanged: `https://cc1234-stashface-onnx.hf.space` and `http://localhost:7860` only.

### DIFF
--- a/plugins/visage/visage.css
+++ b/plugins/visage/visage.css
@@ -244,9 +244,29 @@
 .visage-toolbar-button.visage-scanning {
   opacity: 1;
   color: var(--v-accent);
-  animation: visage-btn-scan-glow 1.2s ease-in-out infinite;
+  animation: visage-btn-scan-glow 0.8s ease-in-out infinite;
   position: relative;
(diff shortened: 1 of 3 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/visage/11785/3

PR #748, opened 2026-07-29
Add List Sort Filter plugin

suzuhiroruri:
Trims the **sort by** dropdown of each list down to the options you actually use. Untick an entry and it stops appearing in that list's menu.

Stash offers 30 sort options for scenes, 23 for performers, 13 for tags, and so on; most libraries only ever use a handful.

**Nothing is hidden until you tick it** — installing the plugin changes nothing on its own.

## Display only

Only the menu entry is hidden; sorting itself is untouched. Saved filters keep working even when they sort by a hidden entry, `?sortby=play_count` URLs keep working, and the dropdown button still shows the current sort by name. Hiding an entry cannot break an existing view.

## No hard-coded sort options

The plugin carries no list of sort values and no labels of its own:

- each list's options come from the filter Stash hands its list components (`PluginApi.patch.before` on `SceneList`, `PerformerList`, …), read as the page renders — no interaction, nothing parsed;
- labels come from formatting Stash's own message IDs through `PluginApi.libraries.Intl`, so the settings page matches the dropdown word for word and follows a UI language change immediately;
- what a list offers is re-read every time you open it, so it cannot go stale.

So when Stash adds, removes or renames a sort option, the plugin needs no update. The only Stash names it knows are the eight list component names, which is API surface rather than data.

## Hiding

Stash renders every sort entry with a `data-value` attribute, so the hiding is generated CSS — no DOM rewriting of the menus, no patching of the dropdown:

```css
.sort-by-select .dropdown-menu:has(.dropdown-item[data-value="ANCHOR"])
                               .dropdown-item[data-value="HIDDEN"] { display: none }
```

`ANCHOR` is a sort value that occurs in that list and in no other; `HIDDEN` is an entry the user unticked.

Lists are keyed by the mode Stash uses (`SCENES`, `PERFORMERS`, …), but a stylesheet cannot know which list a menu belongs to and the DOM does not say — the performer list carries the class `gallery-list`, same as the gallery list, and `/performers/42/scenes` is a scene list on a performer route. So each list's rules are anchored on a value unique to it, computed from the lists seen so far. Values that share a name across lists (`rating`, `path`, `tag_count`, `duration`, …) therefore never leak, and lists embedded in performer/studio/tag pages are covered without any route matching. Browsers without `:has()` fall back to a small MutationObserver that stamps each menu with its mode.

## Settings

Stash's plugin settings are static YAML, so a fixed checkbox per sort option would be exactly the thing that needs maintaining whenever Stash changes one. The plugin declares only a master switch plus two JSON values for its own storage, and draws the per-entry checkboxes itself by patching `PluginSettings` — built from Stash's own `SettingGroup` and `BooleanSetting`, so it inherits the native look and the collapsing rather than reimplementing them. A list appears there once you have opened it.

Its own handful of sentences are translated for every language Stash offers (38 tables, falling back to English per string). Those are best-effort — corrections from native speakers are very welcome.

## Testing

Checked against Stash v0.31.1 (Japanese UI):

- scene, performer, group, tag, studio, gallery and image lists all discovered and filtered as configured; lists left unconfigured show every option;
- rules confirmed not to leak between lists that share sort value names;
- settings panel: labels and list names resolved through Stash's translations, native collapsing, toggling a checkbox updates the stored config and the menu;
- `node validate.js` passes.

### DIFF
--- a/plugins/ListSortFilter/ListSortFilter.yml
+++ b/plugins/ListSortFilter/ListSortFilter.yml
@@ -1,0 +1,27 @@
+name: List Sort Filter
+description: Trims the "sort by" dropdown of every list down to the options you actually use - untick an entry and it stops appearing in that list's menu. Display only, so saved filters and ?sortby= URLs are unaffected. Nothing is hard-coded - each list's sort options are read from the filter Stash hands its list components, and labelled with Stash's own message IDs, so no update is needed when Stash changes its sort options or you change UI language. Lists appear in the settings below once you have opened them.
+version: 1.0.0
+url: https://github.com/stashapp/CommunityScripts/tree/main/plugins/ListSortFilter
+ui:
+  javascript:
+    - ListSortFilter.js
+settings:
+  # Only the master switch is a real setting. The per-entry checkboxes cannot
+  # live here: Stash's plugin settings are static YAML, so listing every sort
+  # option would be exactly what needs maintaining whenever Stash changes one.
+  # ListSortFilter.js draws them instead, from the menus it has discovered.
+  enabled:
+    displayName: Enable this plugin
+    description: Turn off to show every sort option again, in every list. Default on.
+    type: BOOLEAN
+  # Storage, not preferences: the discovered menus and the hidden entries, as
+  # JSON. Kept in the plugin config (rather than the browser) so the choices
+  # follow you across devices. Their rows are hidden by the plugin's own UI.
+  lists:
+    displayName: 'Internal: discovered lists (JSON)'
+    description: Managed by the plugin. Clear it to forget every discovered list; they are re-read next time you open one.
+    type: STRING
+  hidden:
+    displayName: 'Internal: hidden entries (JSON)'
+    description: Managed by the plugin. Clear it to show every sort option again.
+    type: STRING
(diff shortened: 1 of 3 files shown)


plugins/ListSortFilter/ListSortFilter.yml
```diff
@@ -0,0 +1,27 @@
+name: List Sort Filter
+description: Trims the "sort by" dropdown of every list down to the options you actually use - untick an entry and it stops appearing in that list's menu. Display only, so saved filters and ?sortby= URLs are unaffected. Nothing is hard-coded - each list's sort options are read from the filter Stash hands its list components, and labelled with Stash's own message IDs, so no update is needed when Stash changes its sort options or you change UI language. Lists appear in the settings below once you have opened them.
```
DogmaDragon, line 2: Description is too verbose.

plugins/ListSortFilter/README.md
DogmaDragon, line 1: Please reformat to remove mid-sentence line breaks.

DogmaDragon (changes requested):
The low effort submissions are getting tiring. You already had to solve these same complains in your previous PR.

PR #749, opened 2026-07-29
Mosaic Poster: translate the settings page

suzuhiroruri:
Follow-up to #743. The plugin's on-poster strings already follow the UI language, but its **settings page stayed English**: Stash serves plugin settings from the YAML as plain text, and there is no API for a plugin to register translations. This translates them at render time.

### What it does

- **Setting rows** — patches Stash's `PluginSettings` component and hands this plugin's props on with `display_name` / `description` swapped for the UI language. The rows are still rendered by Stash itself; every other plugin's props are passed straight through (`pluginID` guard), and the arity of the patch chain is preserved so a second plugin patching the same component keeps working.
- **The plugin description above the rows, and the two task entries under Settings → Tasks** — neither is drawn from anything a plugin can reach, so those are corrected in the DOM, anchored on this plugin's own rows (`plugin-MosaicPoster-*`) and matched on their English text.
- **Strings live in `MosaicPoster.settings.json`**, served from the plugin's own assets and fetched only when a settings page is opened — settings text in ~40 languages is several times the size of the plugin JS, and no other page shows it.

Same language coverage as the existing `STRINGS` table (40 entries). Lookup tries the full code, then the base code, then English, so an unlisted locale simply stays English. If `window.PluginApi` is unavailable the patch is skipped and nothing changes.

### Verified

In a clean browser profile against Stash v0.31.1:

| UI language | result |
| --- | --- |
| `ja-JP` | all 5 settings, the description and both task entries translated |
| `de-DE` | exact-code match |
| `pt-BR` | falls back to the `pt` entry |
| `en-GB` | English, unchanged |

Other plugins' settings rows (56 of them) are untouched in every language, another plugin that patches `PluginSettings` still renders its own panel, and scene pages log no console errors. `validator` passes.

Non-English strings are best-effort — corrections from native speakers are very welcome.


discourse-stashapp:

https://discourse.stashapp.cc/t/mosaic-poster/12358/2

PR #752, opened 2026-08-07
feat: Gallery covers added to preview

Servbot91:
- Adds gallery covers to thumbnail preview random selection
- Fixes card collapse issue on performer page causing a visual bug or jitter when mouse hover over on certain pieces of the card or when generation takes a while to load. The fix ensures the card always remains its base size before showing preview.


discourse-stashapp:

https://discourse.stashapp.cc/t/thumbnail-previews/1947/4

PR #755, opened 2026-08-10
Update Visage to v2.4.1

cc1234475:
## What's new in v2.4.1

Bug fix release for whole scene scanning.

### Scene scan no longer fails with a play() AbortError

Some users saw `Scene scan failed: The play() request was interrupted because the media was removed from the document.` when a scene had no sprite sheet and the scan fell back to the preview video.

The frame extractor called `play()` on an off-screen video before seeking. That promise can be rejected for reasons outside the plugin's control, and the rejection killed the whole scan. Seeking a `preload="auto"` video already decodes and presents the frame, so playback was never needed. The call is gone.

### Frame extraction hardening

- Seek waits time out after 10s. Before, a stalled decode left the scan spinning forever with no way out except a page reload.
- A stalled or failed seek now keeps the frames already captured and identifies on those, instead of failing the scan.
- Previews reporting a non-finite duration are rejected with a clear message. ffmpeg previews can report `Infinity`, which produced NaN seek times.
- The video element and both canvases are released in a `finally` block, so failed scans no longer leak the element and its buffered data.

### Verified

Local Stash instance, Chrome: 12 distinct frames extracted in 235ms with zero `play()` calls, both the sprite and the preview scan paths complete end to end, and load failures surface a plain message with no unhandled rejections.

### DIFF
--- a/plugins/visage/visage.js
+++ b/plugins/visage/visage.js
@@ -9004,6 +9004,11 @@
         })));
     }
 
+    const LOAD_TIMEOUT_MS = 15000;
+    const SEEK_TIMEOUT_MS = 10000;
+    // Below this delta the browser treats the seek as a no-op and never fires
+    // `seeked`, so the frame is already the one we want.
+    const SEEK_EPSILON = 0.001;
     function pad(n, width = 2) {
         return String(n).padStart(width, '0');
     }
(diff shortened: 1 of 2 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/visage/11785/4

PR #756, opened 2026-08-10
Allow crontab schedule

nocoolnametom:
This allows setting the schedule using crontab format


discourse-stashapp:

https://discourse.stashapp.cc/t/stash-scheduler/7059/3

PR #758, opened 2026-08-14
Add Visage v2.5.0 - first-run backend onboarding + settings UX

cc1234475:
## Visage v2.5.0

Face recognition plugin for Stash.

### Changes
- **First-run backend onboarding**: new-user dialog guiding setup of the face-recognition service.
- **Backend config UI**: badge + gear entry point in the Stash toolbar to configure the Gradio API endpoint from a single source.
- **Poster capture fix**: capture the video frame from its poster image when the video has not loaded, instead of the whole-page fallback.
- **CSP**: ship the wildcard `http://*:7860` connect-src so any local / LAN / NAS host on port 7860 works without per-host config.

This PR contains only the Visage v2.5.0 release changes (`plugins/visage/{visage.js,visage.css,visage.yml}`).

### DIFF
--- a/plugins/visage/visage.yml
+++ b/plugins/visage/visage.yml
@@ -9,7 +9,12 @@ ui:
   - visage.css
   csp:
     connect-src:
+    # Local backend default: http://localhost:7860
+    # If your backend runs on another address (e.g. your LAN IP), add it below
+    # or the browser CSP will block requests. Reapply after every Visage update.
     - "https://cc1234-stashface-onnx.hf.space"
-    - "http://*:7860"
+    - "http://localhost:7860"
+    # Example (uncomment and replace):
+    # - "http://192.168.1.50:7860"
     script-src:
     - "'self'"
--- a/plugins/visage/visage.yml
+++ b/plugins/visage/visage.yml
@@ -1,6 +1,6 @@
 name: Visage
 description: Use facial recognition to lookup performers from scene sprites
-version: 2.4.1
+version: 2.5.0
 url: https://discourse.stashapp.cc/t/visage/11785
 ui:
   javascript:
@@ -10,6 +10,6 @@ ui:
   csp:
     connect-src:
     - "https://cc1234-stashface-onnx.hf.space"
-    - "http://localhost:7860"
+    - "http://*:7860"
     script-src:
     - "'self'"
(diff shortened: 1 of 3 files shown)


plugins/visage/visage.yml
```diff
@@ -10,6 +10,6 @@ ui:
   csp:
     connect-src:
     - "https://cc1234-stashface-onnx.hf.space"
-    - "http://localhost:7860"
+    - "http://*:7860"
```
DogmaDragon: I don't like wildwcard in CSP overrides. It's a bad security policy.

DogmaDragon:
Please note that the commit is made by a different user. Make sure it's not an accident.

plugins/visage/visage.yml
cc1234475: It was the only way I could make it user friendly enough so that people don't have to go modify the Yaml config. Maybe this needs to be a feature request on the stash side so that we can have dynamic CSP.

cc1234475:
user changed on commit.

plugins/visage/visage.yml
DogmaDragon: Isn't this only for local backend which already requires external configuration? I think sane defaults beat the friction of needing to modify the config file manually. 

I do agree that this could be improved on Stash, at the very least allowing to modify the config inside Stash.

discourse-stashapp:

https://discourse.stashapp.cc/t/visage/11785/7

plugins/visage/visage.yml
cc1234475: To some extent, this comes down to a lack of knowledge about how users will actually use it. If people are running things locally, then a sensible default would be localhost. However, I know that a lot of people on the Discord have Stash running all the time on one of their servers. In that case, they would need to specify which server to use by going into the plugin's configuration YAML file and modifying it, which is what I was trying to avoid.

On closer consideration, though, I believe we should go with a more secure default for localhost and accept some friction for power users who are running this on something other than localhost, since they will need to modify the YAML file to point to where their server is running, and again after every update.

I've put in a pull request with what I believe could be a possible solution, allowing us to generate CSPs that could be part of the plugin settings: https://github.com/stashapp/stash/issues/7165. This might be a better long term solution.

PR #766, opened 2026-08-29
BulkImageScrape: enable `force_api_key`

punamust:
During long bulk image scraping tasks the session cookie can time out. This leads to tons of authentication errors and the plugin stops somewhere half-way in the middle.

I expected this to be unnecessary because stashapi is supposed to use the `ApiKey`, but Stash itself actually never passes it: https://github.com/stg-annon/stashapi/issues/34.
Enabling `force_api_key` makes stashapi look up the `ApiKey` using the session cookie and continue using that (if it's generated).

~~I haven't tested yet if this actually works. I should find out soon.~~ It seems to work.

### DIFF
--- a/plugins/bulkImageScrape/BulkImageScrape.yml
+++ b/plugins/bulkImageScrape/BulkImageScrape.yml
@@ -1,7 +1,7 @@
 name: Bulk Image Scrape
 # requires: PythonDepManager 
 description: Apply an image scraper to all images
-version: 0.3.5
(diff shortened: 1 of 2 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/bulk-image-scrape/1339/13

DogmaDragon (changes requested):
Needs to bump the version in the .yml file.

PR #768, opened 2026-08-29
fix(RenameFile): fix doubled file extension causing 'File name too long' error

### LINKED ISSUE #767: RenameFile: OSError 'File name too long' from doubled extension in long-filename truncation
### Plugin
RenameFile

### Description

Renaming a scene whose generated filename exceeds `max_filename_length` (default 255) can produce a filename that is *longer* than the original and ends in a doubled extension (e.g. `....mp4.mp4`), causing the OS rename to fail with `OSError: [Errno 63] File name too long` on macOS (and the equivalent `ENAMETOOLONG` elsewhere).

### Root cause

In `rename_scene()` (`renamefile.py`), the truncate-and-hash fallback appends the file extension onto `new_filename` *before* computing/appending the `_`+MD5-hash suffix, and doesn't reserve space in the length budget for that suffix:

```python
if len(new_filename) > max_filename_length:
    extension_length = len(Path(original_file_path).suffix)
    max_base_filename_length = max_filename_length - extension_length
    truncated_filename = new_filename[:max_base_filename_length]
    hash_suffix = hashlib.md5(new_filename.encode()).hexdigest()
    new_filename = truncated_filename + '_' + hash_suffix + Path(original_file_path).suffix
newFilenameWithExt  = new_filename + Path(original_file_path).suffix
```

Two bugs:
1. `max_base_filename_length` only subtracts the extension length, not the `_` + 32-char hash that gets appended afterward — so the "truncated" name can still exceed `max_filename_length`.
2. Since `new_filename` already ends with the extension after truncation, the next line (`newFilenameWithExt = new_filename + suffix`) appends the extension **a second time**, producing a name like `....mp4.mp4` and pushing it well past the OS limit.

### Reproduction (from a real log)

```
OSError: [Errno 63] File name too long: '.../Movie-name-(...)-[...].mp4' -> '.../Movie-name-(...)-[...Gagging_a68f3c2a62cef848afd6bb9ef0ff4fc6.mp4.mp4'
```
The resulting target filename is 292 characters (vs. the configured 255 limit) and ends in `.mp4.mp4`.

### Additional issue

When the rename/move fails, the code logs a clear one-line error (`stash.Error(exitMsg)`) but then re-raises the `OSError`, which propagates to the plugin's top-level catch-all and dumps the full Python traceback to the log as well — burying the clear message under noise.

---
*Disclosure: this issue and the accompanying fix (PR to follow) were drafted with LLM (Claude) assistance, based on analysis of a real error log; reviewed and submitted by a human maintainer.*

garfolino:
Fixes #767

**Independent of #733.** This branch was previously stacked on `fix/renamefile-title-stuck`; it has been rebased onto `main` and now contains only the `renamefile.py` changes described below. The two PRs no longer overlap and can be reviewed separately. Note the version number assumes #733 lands first — if this merges first, the bump here should be `1.0.2` instead.

### Summary
- Fix the long-filename truncation fallback in `RenameFile` so it no longer doubles the file extension and correctly reserves space for the hash suffix, which was causing `OSError: [Errno 63] File name too long` (macOS) / `ENAMETOOLONG` when renaming scenes whose formatted filename exceeded `max_filename_length`.
- When a rename/move does fail, stop the OSError from being re-raised into the top-level catch-all (which dumped a full Python traceback into the log on top of the already-logged clear error message).
- Bump plugin version to 1.0.3 and update the changelog.

### Root cause
See #767 for the full analysis. In short: the truncation branch built `new_filename` as `truncated + '_' + hash + extension`, but the length budget it truncated to only subtracted the extension length (not the `_`+hash), and the caller then appended the extension a second time (`newFilenameWithExt = new_filename + suffix`), producing filenames like `....mp4.mp4` well past `max_filename_length`.

### Testing
- Reproduced the exact reported failure by hand: recomputed the original 292-character doubled-extension filename from the logged error and confirmed the new logic instead produces a name at or under the configured `max_filename_length`, with a single extension.
- Ran the repo's plugin validator (`node ./validator/index.js --ci`) — passes.
- Not yet exercised against a live Stash instance with an actual over-long title/tag combination; recommend a maintainer or the reporter re-test the original scenario before merge.

### AI disclosure
This fix (analysis, code changes, issue text, and this PR description) was drafted with LLM (Claude) assistance from a real user-reported error log. The diff was reviewed by a human before submission, per the repo's LLM-assisted contribution policy.

### DIFF
--- a/plugins/RenameFile/renamefile.py
+++ b/plugins/RenameFile/renamefile.py
@@ -388,12 +388,12 @@ def rename_scene(scene_id):
     original_file_name  = Path(original_file_path).name
     new_filename        = form_filename(original_file_stem, scene_details)  
     max_filename_length = int(config["max_filename_length"])
-    if len(new_filename) > max_filename_length:
-        extension_length = len(Path(original_file_path).suffix)
-        max_base_filename_length = max_filename_length - extension_length
-        truncated_filename = new_filename[:max_base_filename_length]
+    extension_length = len(Path(original_file_path).suffix)
+    if len(new_filename) + extension_length > max_filename_length:
         hash_suffix = hashlib.md5(new_filename.encode()).hexdigest()
-        new_filename = truncated_filename + '_' + hash_suffix + Path(original_file_path).suffix
+        max_base_filename_length = max_filename_length - extension_length - len(hash_suffix) - 1
+        truncated_filename = new_filename[:max_base_filename_length]
+        new_filename = truncated_filename + '_' + hash_suffix
     newFilenameWithExt  = new_filename + Path(original_file_path).suffix
     new_file_path       = f"{original_parent_directory}{os.sep}{new_filename}{Path(original_file_name).suffix}"
     org_file_root_stem  = f"{original_parent_directory}{os.sep}{original_file_stem}"
--- a/plugins/RenameFile/renamefile.py
+++ b/plugins/RenameFile/renamefile.py
@@ -444,15 +444,15 @@ def rename_scene(scene_id):
             stash.Trace(f"Calling [metadata_scan] for path {original_parent_directory.resolve().as_posix()}")
             stash.metadata_scan(paths=[original_parent_directory.resolve().as_posix()])
         if targetDidExist:
-            raise
+            return None
         if os.path.isfile(new_file_path):
             if os.path.isfile(original_file_path):
                 os.remove(original_file_path)
             pass
         else:
             # ToDo: Add delay rename here
-            raise
-    
+            return None
+
     if dry_run:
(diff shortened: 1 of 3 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/renamefile/1334/9

DogmaDragon (changes requested):
Need to resolve the conflicts.

dude-stash:
@DogmaDragon given that [David-Maisonave](https://discourse.stashapp.cc/u/david-maisonave) has not been seen for almost a year, can we go ahead and merge these? This one was done on top of #733 so I may need to change the readme here before it merges, assuming it merges after

PR #769, opened 2026-08-30
feat(sceneRename): rename via moveFiles and record original filename

stashdbcorrode248:
SceneRename renamed files with `os.rename()` behind Stash's back, then ran `metadata_scan()` on the parent folder to repair the database afterwards. The file record pointed at a stale path until that scan completed, and a full folder scan is heavy for a single-file change.

This switches the rename to the `moveFiles` mutation, so Stash moves the file and updates the file record in one transaction and no rescan is needed. A destination folder is required even for an in-place rename, so the file's current `parent_folder` id is passed, falling back to its path.

### Other changes

- **Require a title, not just a studio.** Without one the filename collapsed to the studio plus resolution, which is less useful than the original name. The README already documented this requirement; the code now matches it. The skip message names the missing field rather than listing every field.
- **Record the pre-rename basename** in the scene's `original_filename` custom field, written once so the earliest known name survives later renames. A failed write logs a warning and does not fail the rename.
- **Harden filename sanitising.** `:` is now stripped from studio names and codes, which never passed through `clean_title()`; control characters are collapsed to spaces; leading dots and trailing dots/spaces are trimmed.
- **Truncate to a byte budget rather than a character count.** Filesystem name limits are in bytes (255 on ext4/btrfs), so a 240-character non-ASCII title could produce a name well over the limit and fail to rename. The budget also reserves room for the extension and a duplicate suffix.
- **Correct the README** format string and worked example, which showed the resolution before the title while `form_filename()` emits it after.

Version bumped to 1.2.0.

### Testing

Run against a live Stash 0.31.1 library. Renames land through `moveFiles` with no rescan and no stale paths; `original_filename` is populated; scenes missing a title are skipped with the field named in the log. Sanitising was checked against studio/title combinations containing colons, newlines, tabs, NUL, leading dots and trailing dots/spaces, plus a 300-character CJK title, which previously produced a 706-byte name and now stays within the 255-byte limit.

### DIFF
--- a/plugins/sceneRename/README.md
+++ b/plugins/sceneRename/README.md
@@ -10,6 +10,9 @@ Simple plugin to help organize scene files into a clean, consistent format. It i
 * Graceful handling of already-renamed files
 * Does not fail if Scene ID or resolution are missing
 * Requires a Studio and Title to proceed
+* Renames through Stash's `moveFiles` API, so the database stays in sync without a rescan
+* Records the pre-rename filename in the scene's `original_filename` custom field
+* Strips illegal and control characters, and truncates to the filesystem's byte limit
 
 The code is simple and the plugin UI includes clear usage instructions.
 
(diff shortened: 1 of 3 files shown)


discourse-stashapp:

https://discourse.stashapp.cc/t/scenerename/5795/4

PR #774, opened 2026-09-07
fix(javstashAutofill): correct name casing, store the stash-box id, simplify the source setting

### LINKED ISSUE #773: [Javstash Autofill] Scraped name never replaces the typed name; stash-box id is not stored; source needs a URL Stash already has
Plugin: Javstash Autofill v1.3 (`plugins/javstashAutofill/javstash_autofill.py`)

Three separate problems. I have a fix for all three and will open a PR.

### 1. The scraper's name never replaces the created name

With *use scraper name as primary* enabled, a performer whose name I typed in lowercase keeps the lowercase name. The rename is gated on:

```python
prefer_scraper = bool(use_scraper_name) and bool(cand_name) and norm(cand_name) != norm(name)
```

`norm()` lowercases and strips spaces, so `yua mikami` and `Yua Mikami` compare equal and the rename is skipped. Only a genuinely different name triggers it — a casing or spacing difference, which is the common case when typing a name by hand, never does.

Expected: the primary name is corrected to the scraper's spelling. A casing-only correction is arguably safe even with the toggle off, since it is the same name.

### 2. The stash-box id is not stored

When the source is a stash-box, the matched performer's remote id is never written to `stash_ids`. The scrape query does not ask for it:

```python
JAV_FIELDS = ("name gender birthdate death_date ethnicity country hair_color eye_color "
              "height weight measurements fake_tits career_length tattoos piercings "
              "details twitter instagram aliases images")
```

`remote_site_id` is missing, so the id is not available to the update at all. After autofill the performer still has no stash id, and Identify / the Tagger do not recognise it afterwards.

Related, same query: `build_update()` reads `cand.get("url")`, but neither `url` nor the modern `urls` list is requested, so scraped URLs other than twitter/instagram are always dropped.

### 3. Configuring the source is harder than it needs to be

`Scraper (Identify)` and `Scraper (manual)` want a full stash-box endpoint URL, which Stash already knows from Settings → Metadata Providers:

```python
def build_source(source_str):
    # A URL is treated as a stash-box endpoint; anything else as a scraper_id.
    s = (source_str or JAV).strip()
    return {"stash_box_endpoint": s} if s.startswith("http") else {"scraper_id": s}
```

Retyping the URL is also a silent failure mode: `resolveStashBox` in stash only accepts an endpoint matching a configured one exactly (case-insensitively), so a trailing slash or a slightly wrong path fails with "stash box not found" and the hook only logs a scrape error.

Proposal: resolve the setting against `configuration.general.stashBoxes` and `listScrapers(types: [PERFORMER])`, so the value may be a stash-box name (`StashDB`), part of a name or endpoint (`stashdb`), a performer scraper name or id, or still a full URL. Empty falls back to javstash when configured, otherwise the first configured stash-box.

### Steps to reproduce (1 and 2)

1. Configure a stash-box (e.g. StashDB) under Settings → Metadata Providers and set it as the plugin's manual source.
2. Turn on *Manual - use scraper name as primary*.
3. Create a performer manually, typing the name in lowercase (`yua mikami`).
4. The fields are filled from the scraper, but the name stays `yua mikami` and no stash id is added.

dude-stash:
Fixes #773. Three fixes for Javstash Autofill, version bumped to 1.4.

### 1. The name is now corrected to the scraper's spelling

I usually type performer names without the right casing, and the plugin never fixed them. The rename check was:

```python
prefer_scraper = bool(use_scraper_name) and bool(cand_name) and norm(cand_name) != norm(name)
```

`norm()` lowercases and strips spaces, so `yua mikami` and `Yua Mikami` are equal there and nothing happens.

Now:

```python
same_name = norm(cand_name) == norm(name)
prefer_scraper = (bool(cand_name) and nfc(cand_name) != nfc(name)
                  and (bool(use_scraper_name) or same_name))
```

A difference in casing or spacing is always corrected, because it is the same name. A really different name still needs the "use scraper name as primary" toggle, as before. `find_by_name()` compares with `norm()` now as well, so an existing "Yua Mikami" is found as the duplicate and merged, instead of the rename hitting it.

### 2. The stash-box id is stored

`remote_site_id` was missing from the scrape query, so the id was never available. It is requested now and, when the source is a stash-box, written into `stash_ids` in all three paths (fill, rename, merge), without touching ids from other endpoints. Identify and the Tagger recognise the performer after autofill.

`urls` was added to the query too, because `build_update()` was reading `cand.get("url")` and that field was never fetched.

### 3. The source setting no longer needs the endpoint URL

Before, the setting wanted the full endpoint URL that Stash already has under Settings → Metadata Providers. Worse, `resolveStashBox` only accepts an endpoint that matches a configured one exactly, so a trailing slash or a small typo failed with "stash box not found" and the hook only logged a scrape error.

`resolve_source()` now looks the value up in `configuration.general.stashBoxes` and `listScrapers(types: [PERFORMER])`, in this order:

1. stash-box name, e.g. `StashDB`
2. exact endpoint URL
3. any URL, passed through as before
4. part of a name or endpoint, e.g. `stashdb`
5. performer scraper id or name, e.g. `Minnano-AV-JP`

Empty uses javstash if it is configured, otherwise the first configured stash-box. The log line now says which source was used, for example `via stash-box 'StashDB' (by name)`.

Old values keep working: a full URL and a `scraper_id` are still accepted, so nobody has to change their settings.

### Testing

- Tested against my own library: a manually created performer typed in lowercase gets the correct name, the StashDB id in `stash_ids`, and the fields filled.
- The source resolution and the update building were also checked with a stubbed GraphQL client for every case listed above, including a stash-box that is not configured and a scraper name that does not exist.
- `node validate.js plugins/javstashAutofill/javstashAutofill.yml` passes.

README and CHANGELOG updated.

AI assistance was used for this change.

### DIFF
--- a/plugins/javstashAutofill/CHANGELOG.md
+++ b/plugins/javstashAutofill/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.4
+- Fix: the primary name is now corrected to the scraper's spelling when it only differed in casing/spacing (a typed `yua mikami` becomes `Yua Mikami`); previously such a name was considered identical and never replaced.
+- Fix: the matched stash-box performer id is stored as a `stash_id` (the scrape now requests `remote_site_id`), so Identify and the Tagger recognise the performer afterwards.
+- Easier configuration: the source can be given as the name of a stash-box already configured in Stash (e.g. `StashDB`), part of a name/endpoint (`stashdb`), or a performer scraper name - no need to retype the endpoint URL. Empty falls back to javstash if configured, otherwise the first configured stash-box.
+- Scraped `urls` are now read from the modern `urls` list as well as twitter/instagram.
+
 ## v1.3
 - Initial release.
 - Fill a newly-created performer's empty fields from a scraper (default javstash) on `Performer.Create.Post`.
--- a/plugins/javstashAutofill/README.md
+++ b/plugins/javstashAutofill/README.md
@@ -10,8 +10,9 @@ It searches the configured source by the performer's name, takes the best name-m
 
 - **Best name match, or skip.** The candidate with the highest name-match score (exact = 1.0, otherwise fuzzy) is used. If the best score is below the configurable threshold, nothing is done.
 - **Fill empty only (default).** Only empty fields are written, so manual edits and other scrapers' data are preserved. Each field can be switched to "overwrite existing" individually.
-- **Primary name policy.** For each origin (Identify vs. manual creation) you can choose whether the created name stays primary (scraper name goes to aliases) or the scraper's canonical name becomes primary (created name goes to aliases).
+- **Primary name policy.** For each origin (Identify vs. manual creation) you can choose whether the created name stays primary (scraper name goes to aliases) or the scraper's canonical name becomes primary (created name goes to aliases). Either way, a name that only differs in casing or spacing is always corrected to the scraper's spelling — typing `yua mikami` gives you `Yua Mikami`.
 - **Duplicate merge.** When the scraper name would become primary but a performer with that name already exists, the newly-created one is merged into the existing performer (`performerMerge`) — scenes are re-linked, the created name and the stash-box id are carried over, and the duplicate is removed.
+- **Stash-box id stored.** When the source is a stash-box, the matched performer's remote id is saved as a `stash_id`, so Identify and the Tagger recognise the performer from then on.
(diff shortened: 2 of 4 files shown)


garfolino:
@suzuhiroruri Love the plugin, its been very helpful. I added a few improvements. :)

discourse-stashapp:

https://discourse.stashapp.cc/t/javstash-autofill/11502/2

PR #775, opened 2026-09-10
stashdb-performer-gallery: fix infinite loop in relink_images when performer has >100 images

adtdev20:
Page number and Item counter were being set to the same variable, so when a performer had more than 100 images, the loop would jump from page 1 to page 100. With page 100 not having any content, it would drop back to 0, and read 0+1, so page 1 again, and just get stuck there indefinitely.

The code in tagImagesFromGalleries correctly handled the count + pagination so that's the template for this fix.


discourse-stashapp:

https://discourse.stashapp.cc/t/stashdb-performer-gallery/1411/17

DogmaDragon (changes requested):
Need to bump the version in the .yml file.

adtdev20:
> Need to bump the version in the .yml file.

Done!

Tweeticoats:
looks good

